Issue #402, opened 2026-06-21
Perf: micro-optimizations in injection range/offset helpers


atusy:
## Context
Surfaced by Gemini during the review of #400; deferred because that PR was a pure structural relocation (verbatim move) and these are behavioral/perf changes.

## Items
- [ ] `content.rs` `compute_line_column_offsets`: the inner `for line in gap_start_line..=gap_end_line` loop only does work when `line == gap_start_line`; the rest just resizes. Resize once to `gap_end_line + 1` and set the offset directly. Optionally add a fast O(1) content-node bounds check before the gap computation to skip regions that can't contain the target.
- [ ] `ranges.rs` `compute_included_ranges_in_window`: the `has_nonzero_children` pre-check traverses all named children once before the main loop. Consider folding the "all children zero-width / single trivial range" case into the final filter to avoid the extra pass (verify equivalence with the zero-width `block_continuation` ghost-node case first).
- [ ] `discovery.rs` `collect_all_injections`: `injections_map.entry(key).or_insert(InjectionRegionInfo { ... })` eagerly evaluates `has_include_children_for_pattern` / `effective_offset_for_pattern` (both iterate query predicates) even for duplicate content-node ranges. Switch to `or_insert_with`.
- [ ] (evaluate — currently documented as intentional) `ranges.rs` `sub_select_included_ranges`: Gemini argued that when the first clipped range is on a single row, `end_point.column` should be reduced by the same delta as `start_point.column` (which is zeroed). The current behavior is deliberate (doc comment "Known limitation"; tree-sitter validates byte ordering only; first-range `column = 0` avoids double-counting the `> ` prefix) and is covered by a passing test. Confirm whether this actually produces wrong columns in any real blockquote-injection rendering before changing it.

Refs: #400 review threads.

atusy:
Resolved by #411 (merged, `b9f1b89cb`). Outcome per item:

- [x] **`compute_line_column_offsets` resize-once** — implemented (hoisted the per-row `resize` to a single `resize(gap_end_line + 1, 0)`; equivalent because gaps arrive in ascending, non-overlapping document order). The *optional* "O(1) content-node bounds check" addendum was **not** added — it targets a position-lookup fast path, whereas `collect_all_injections` / `compute_line_column_offsets` run over the whole document, so it doesn't apply here.
[...]

Issue #401, opened 2026-06-21
Harden injection read-path string slicing against stale-tree/text desync


atusy:
## Context
Surfaced by Gemini during the review of #400 (a pure structural relocation that moved this code verbatim from `main`, so the hardening was deliberately deferred to keep that PR behavioral-change-free).

Several injection read-path helpers slice `&text[a..b]` / `text[..n]` directly with byte offsets taken from tree-sitter node ranges. If the `tree` and `text` ever desync (stale tree, concurrent edit), those offsets can be out of bounds or land mid-codepoint, which **panics and crashes the LSP server**. This repo has prior crash history of exactly this class (e.g. #348 unedited-tree reparse, didChange concurrency races), so it's worth hardening rather than relying on the "tree and text come from one `DocumentSnapshot`" invariant holding everywhere.

## Sites (all in `src/language/injection/`)
- `discovery.rs` — `position_of_byte` (`text[lo..hi]`, `text[..byte_pos]`, `text[line_start..byte_pos]`)
- `discovery.rs` — `CacheableInjectionRegion::from_region_info` (the `None` offset branch `text[start_byte..end_byte]`; the `start_byte == node.start_byte()` branch with `node.start_byte() - byte_column`)
- `content.rs` — `byte_to_point_anchored` (`&text[anchor_byte..clamped]`)
- `content.rs` — `extract_clean_content` (`host_text[byte_range]`, `content[range.start_byte..range.end_byte]`)
- `content.rs` — `compute_line_column_offsets` (`gap_abs_byte - range.start_point.column`, `host_text[line_start_byte..gap_abs_byte]`)
- `language.rs` — `extract_dynamic_language`, `extract_language_from_info_string` (`text[capture.node.byte_range()]`)

## Suggested approach
Use `str::get(a..b).unwrap_or("")` (degrade gracefully to empty) and `saturating_sub`; snap offsets to char boundaries with `crate::text::{ceil_char_boundary, floor_char_boundary}` (ceil the start, floor the end so the offset window never re-includes excluded bytes). This is a **behavioral change** (it alters out-of-bounds handling), so it needs its own PR with regression tests that feed deliberately-stale offsets.

Refs: #400 review threads.

PR #411, opened 2026-06-22
perf(injection): micro-optimizations in range/offset helpers (#402)

atusy:
Implements #402.

> **Stacked on #407 (#401).** Base is `fix/401-injection-slice-hardening`; this PR's diff is only the perf delta. It will retarget to `main` automatically once #407 merges. The `compute_line_column_offsets` change builds on the hardened version from #401, which is why it's stacked rather than parallel.

## Implemented

1. **`collect_all_injections`: `or_insert` → `or_insert_with`** — when several injection patterns match the same content-node range, only the first insertion is kept. Deferring the `InjectionRegionInfo` construction skips the `has_include_children_for_pattern` / `effective_offset_for_pattern` predicate scans for the discarded duplicates. Same resulting map.
2. **`compute_line_column_offsets`: hoist the per-row resize** — the inner `for line in gap_start_line..=gap_end_line { offsets.resize(line+1,0); … }` only ever wrote the gap's start row, so it collapses to a single `resize(gap_end_line+1, 0)`. Gaps arrive in ascending, non-overlapping document order (built by `compute_included_ranges_in_window`'s forward walk), so the resize only grows — equivalent output, fewer resize calls.

Both are behavior-preserving for the actual inputs; valid behavior is unchanged.

## Evaluated and intentionally **not** changed

The issue listed two more items; after analysis both are declined:

- **`compute_included_ranges_in_window` pre-check removal** — the suggested "filter a trivial whole-window range to `None` at the end" is **not behavior-preserving**: when non-zero children all fall outside the `#offset!`-clipped window, the loop emits `Some([whole-window range])`, and `compute_line_column_offsets` returns different output for `None` (`vec![start_column]`) vs `Some([range])` (per-line offsets). The existing pre-check is also already an `.any()` short-circuit (O(1) common case), so there's little to gain.
- **`sub_select_included_ranges` first-range `end_point.column`** — this is the function's already-documented "Known limitation" (clipped ranges adjust bytes, not points; tree-sitter validates byte ordering only; the first-range `column = 0` deliberately avoids double-counting the `> ` prefix, and is locked by a passing test). It's pre-existing, not a regression, and untouched by this PR.

## Verification
- `make lint` (clippy `--all-targets --all-features -D warnings`) clean; all unit tests pass; e2e passed via pre-commit hooks.
- Reviewed by subagent multi-perspective review and codex — both confirmed the two changes are behavior-preserving and that items 2 & 4 were correctly declined.

### DIFF
--- a/src/language/injection/content.rs
+++ b/src/language/injection/content.rs
@@ -58,30 +58,29 @@ pub(super) fn compute_line_column_offsets(
             let mut offsets = Vec::new();
             for range in ranges {
                 let gap_start_line = range.start_point.row;
-                // How many virtual lines does this gap span?
                 let gap_end_line = range.end_point.row;
-                for line in gap_start_line..=gap_end_line {
-                    offsets.resize(line + 1, 0);
-                    if line == gap_start_line {
-                        // The column offset is the gap's start column within the host line.
-                        // For the first line (row 0) of the content node, use start_column
-                        // (which accounts for the full prefix from the line start).
-                        // For subsequent lines, the start_point.column from
-                        // compute_included_ranges is already the raw byte column
-                        // within the line — convert to UTF-16.
-                        if line == 0 {
-                            offsets[line] = start_column;
-                        } else {
-                            // Convert byte column to UTF-16 code units.
-                            // The gap's start_byte is relative to content node start.
-                            // The line starts at the byte after the previous newline.
-                            let gap_abs_byte = byte_range.start.saturating_add(range.start_byte);
-                            let line_start_byte =
-                                gap_abs_byte.saturating_sub(range.start_point.column);
-                            let prefix = clamped_slice(host_text, line_start_byte..gap_abs_byte);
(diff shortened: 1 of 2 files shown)


atusy:
### Review status — converged

- **Subagent multi-perspective review**: safe; verified both optimizations are behavior-preserving and that the two declined items (the `compute_included_ranges_in_window` post-filter and the `sub_select_included_ranges` column) were correctly not implemented.
- **Codex**: no findings; agreed the `or_insert_with` change has no LSP-observable consequence (the only changed observable is an `info` log for a *duplicate*-pattern malformed `#offset!`, whose parsed value is discarded by first-writer-wins anyway).
- **Copilot**: "reviewed 5 out of 5 changed files and generated no comments."
- **Gemini**: rate-limited (daily quota, carried over from #407's review rounds), so it couldn't run this pass — happy to re-request `/gemini review` once the quota resets if you'd like its pass before merging.

0 unresolved threads.

Issue #480, opened 2026-06-24
Move slow didOpen work (auto-install) out of the writer ticket into a spawned task


atusy:
Follow-up to #374. As of #374 `didOpen` is gated as a per-document **writer** in `IngressOrderGate`, which closes the open→edit and close→reopen first-poll-order races. The gate holds the ticket for the whole `did_open_impl` handler future.

## Problem (liveness, not correctness)
Most of the open path is fast or fire-and-forget, but one branch is slow: when the main language's parser is missing and auto-install is enabled, `did_open_impl` awaits `maybe_auto_install_language` → `try_install` (a network/disk download) **inside the handler**, so the writer ticket is held for the whole download. A concurrent same-document `didChange` (and any gated reader) waits that long. This is bounded (first open of an uninstalled parser, one-time) and recoverable, so #374 shipped the minimal gate and deferred this.

The deadlock that historically kept `didOpen` ungated is **not** the issue here — it was stale code-drift (downstream spawn is fire-and-forget; every editor-facing await on the open path is a one-way notification). Verified in #374 via the synthetic-push e2e suite.

## Scope
Restructure `did_open_impl` so the writer-ticketed critical section completes quickly:
- Keep the synchronous document insert + `parse_document` in the handler (a gated `didChange` needs the base tree; readers tolerate insert-without-tree via `wait_for_parse_completion`).
- Move the slow / client-independent tail (auto-install, and optionally `process_injections` orchestration) into a spawned task.
- **Resolve the `skip_parse` coupling deliberately**: today parse is skipped when main-language auto-install fires and `reload_language_after_install` re-parses later. Decide parse's fate when auto-install moves out.
- **Cancel/supersede the spawned task on `didClose`** to avoid a close-races-spawn resurrection (a mid-install task re-inserting a closed doc via `reload_language_after_install`). Reuse the existing eager-open supersede machinery rather than rolling new lifecycle.

Lower priority — current behaviour is correct and pairing-invariant-compliant; this is a responsiveness improvement.

Ref: `src/lsp/ingress_order.rs`, `src/lsp/lsp_impl/text_document/did_open.rs`, `src/lsp/lsp_impl/coordinator/install.rs`.

atusy:
[...]

PR #527, opened 2026-06-29
feat(parse): flip didOpen present-parser parse off the ingress ticket (#6)

atusy:
## Summary

Follow-up **#6** from the per-document-parse-scheduler ADR: flip the `didOpen` present-parser tree-sitter parse **off the per-URI ingress writer ticket**, so a slow/large open parse no longer holds the ticket and wedges later same-URI readers/writers. This mirrors the already-shipped `didChange` off-ingress flip (`reparse_latest`/`schedule_reparse`) and the auto-install spawn (#480/#511). It is the last substantive epoch surface from that ADR.

### What changed
- **`parse_document` is now off-ingress-safe** (`refactor` commit, behaviour-identical while still inline):
  - new non-inserting store CAS `set_parse_result_if_text_and_incarnation_unchanged` — sets the content-refined language **+** tree, guarded by **text** (via `Arc::ptr_eq`, O(1)) **+** **incarnation**; replaces the unguarded `set_parse_result_if_present` (removed) on all three store paths;
  - `populate_injections` gated behind the CAS landing (CAS-before-populate);
  - `advance_watermark_for_incarnation` instead of a plain advance; the doc-gone path returns **without** advancing (per-lifetime watermark);
  - returns whether **this** call's CAS landed the tree.
- **The flip itself** (`feat` commit): `did_open_impl` is a 3-way branch —
  - **auto-install** (`skip_parse`): unchanged (the install spawn already reparses off-ingress);
  - **CLI one-shot**: keeps the **inline** parse — the caller reads the result + the eager-open handles `process_injections` registers the moment `did_open_impl` returns, and there is no concurrent same-URI op to wedge;
  - **interactive LSP**: **spawns** the parse; the handler returns immediately, the spawned parse advances the watermark (releasing a gated reader) and runs the tree-dependent downstream (`process_injections(false)`, the deferred `SemanticTokensRefresh`, the synthetic diagnostic) **only when this parse's CAS landed the tree**.
- **Open CAS uses `Arc::ptr_eq`** (`perf` commit) — removes the only O(n) compare the safety refactor added on the large-document open path; plus a cold-open latency benchmark scenario.
- **Downstream gated on the parse's own CAS, not "a tree exists"** (`fix` commit, codex): when a racing `didChange` reparse wins the CAS, the open parse runs no downstream — the edit reparse owns the tree and already ran `process_injections(forward=true)`.

### Performance (the hard constraint: no degradation)
A/B benchmark (HEAD off-ingress vs `origin/main` inline, `didOpen` → first `semanticTokens`):

| scenario | main (inline) | HEAD (off-ingress) | Δ |
|---|---|---|---|
| markdown 150-block open | 49.0ms | 43.2ms | **−11.8%** |
| rust 1500-func open (host parse < budget) | ~neutral | | |
| rust 4000-func open (~150KB; host parse > 200ms, on-demand fallback fires every run) | ~1476ms | ~1472ms | **~0%** |

HEAD is faster or neutral everywhere — the watermark gates on the **host parse** only, so the expensive `process_injections`/handler-tail leaves the reader's critical path. The on-demand fallback (the only double-parse path) fires solely for ~150KB single dense files, and even there latency is unchanged (token computation dominates; the redundant parse overlaps the off-ingress one). New `Kind::OpenFirstToken` benchmark scenarios validate this regime explicitly.

### Tests
- 2177 lib + full e2e (1470 tests / 0 fail).
- 3 `did_open` unit tests that asserted a synchronous open parse now wait for the async result; the host-attach test's premise inverts (returning before the parse completes is now correct).

### Reviews (all converged)
- **`/review` subagent** — 2 HIGH (the 200ms-budget perf risk + a benchmark gap) both resolved empirically (no latency regression even in the on-demand-fallback regime; the markdown "over budget" scenario was mislabeled — its host parse stays under budget); 1 LOW `Arc::ptr_eq` applied to the open CAS.
- **codex** — 1 HIGH: the open downstream was gated on a `tree().is_some()` re-check instead of the parse's own CAS, so a losing open parse could re-run `process_injections(forward=false)` over a racing edit reparse's tree and supersede its eager-open batch. Fixed by returning a `landed` bool and gating on it.
- **pi-review** — 3 LOW (parse-failure nulled the detected language; `mark_parse_finished` get-or-inserted a parse-state for a closed URI; the auto-install spawn re-fired the synthetic diagnostic in CLI mode), all fixed; confirmation pass clean. Core CAS / watermark / resurrection-safety / Send-lifetime verified sound.

### Known follow-up (out of scope, separate PR)
The auto-install spawn (`reparse_installed_document`) has the same pre-existing "gate on has-tree instead of own CAS" pattern codex flagged for the present-parser path; it predates this change and the fix mirrors this one (return whether it landed the tree).

### DIFF
--- a/benches/semantic_tokens.rs
+++ b/benches/semantic_tokens.rs
@@ -391,6 +391,12 @@ enum Kind {
     /// so each measured request pays incremental reparse + cache invalidation +
     /// delta diff on top of the (full) token recompute.
     EditDelta,
+    /// Cold-open latency: each iteration opens a FRESH document and times
+    /// `didOpen` → first `semanticTokens/full` response — the editor-visible
+    /// "open the file, see highlights" latency. The one scenario that captures
+    /// the open parse itself (no settle sleep), so it is what validates the
+    /// off-ingress open flip (#6) against the reader's watermark settle budget.
+    OpenFirstToken,
 }
 
 /// Mutable per-run state for [`Kind::EditDelta`]: the next `didChange` version,
@@ -426,6 +432,9 @@ fn measure(
     iters: usize,
     warmup: usize,
 ) -> Vec<Duration> {
+    if let Kind::OpenFirstToken = scn.kind {
+        return measure_open(bin, scn, data_dir, iters, warmup);
+    }
     let mut server = Server::start(&bin.path, data_dir);
     server.did_open(scn.uri, scn.language_id, &scn.content);
     // Let the initial parse settle (didOpen parse may be async).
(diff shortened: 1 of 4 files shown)


atusy:
Thanks @gemini-code-assist — applied in ab5e2cf1: `set_parse_result_if_text_and_incarnation_unchanged` now takes `language: Option<&str>` and defers `String::from` to inside the successful CAS branch, so a CAS that rejects (a didChange/reopen racing the off-ingress open parse) allocates nothing; the tree path also drops its `language_name.clone()`. All call sites + tests updated to pass `Some(&language_name)` / `Some("rust")`. Resolving the related threads.

Issue #529, opened 2026-06-30
perf: wire the dead InjectionTokenCache (region-relative tokens + rebase-on-splice)


atusy:
## Summary

The per-`(URI, region_id)` `InjectionTokenCache` (`src/analysis/semantic_cache.rs:176-231`) is **dead code**: its `store`/`get` are `#[cfg(test)]`, and production only ever calls `remove`/`clear_document` (`src/lsp/cache.rs:62,168,180,228,245`). So the invalidation bookkeeping is maintained on every edit, but nothing ever populates or reads it — injection tokens are **recomputed on every `semanticTokens` request** (`src/analysis/semantic/parallel.rs:703-739`, parallelized across Rayon workers but still full CPU work).

Wiring it in would let an edit that lies *outside* a region reuse that region's cached tokens instead of re-tokenizing it.

## Why this is NOT a quick win (honest scoping)

Discovered during the 2026-06-30 post-parse-scheduler perf sweep (4-lens ultrathink + advisor):

- **The blocker is coordinate-remapping geometry, not invalidation.** Injection tokens are emitted in **host-absolute `(line, column)`** coordinates and merged with host tokens into one sorted, delta-encoded stream (`parallel.rs:228-303,742`). A cached region's tokens are only reusable if its absolute start hasn't moved — but a pure positional **shift** of an unchanged region (text inserted *above* it) changes every cached token's absolute line/column and the leading delta encoding. `content_hash`+ULID (already wired, `cache.rs:222-248`) detect *content* edits, not shifts. To reuse the cache you must store **region-relative** tokens and **rebase-on-splice** at read time.
- **The parse-scheduler epoch does NOT unlock this.** Invalidation already works via `content_hash`/ULID; epoch adds nothing to the geometry problem. (This corrects an earlier internal note that called this the "highest-impact perf lever".)
- **Conditional benefit.** Helps only injection-heavy documents (Markdown / Vue / embedded) on localized edits. Does nothing for injection-free documents (plain Rust/Lua — likely the majority), where the host-language recompute dominates and this cache never has entries. And because injection tokenization is already Rayon-parallel, it cuts CPU *work* but only cuts wall-clock latency when injection tokenization is the long pole.

## Effort / risk

- **Effort: HIGH** — region-relative token storage + a rebase-on-splice read/splice path (the coordinate-remapping work).
- **Risk: medium** — subtle wrong-highlight bugs on shifted regions if the rebase is wrong.
[...]

PR #530, opened 2026-06-30
perf(semantic): skip re-tokenizing an unchanged document (content+generation keyed cache)

atusy:
## What

`semanticTokens/full` and `full/delta` recomputed the **entire** token set on every request: the `SemanticTokenCache` was keyed only by `result_id`, a fresh global counter per response, so it never matched the document's current state and never hit. The change tags cached tokens with a `cache_key` = FNV-1a hash of the text **folded with a settings generation**; before tokenizing, both handlers check `get_current_tokens(cache_key)` and on a hit (unchanged document, unchanged settings) serve the cached tokens — `full` returns them with their original `result_id`; `delta` still runs its cheap diff against them.

This came out of a post-parse-scheduler perf sweep. The big-ticket lever (wiring the dead `InjectionTokenCache`) is **not** unlocked by the epoch — it's blocked by coordinate-remapping geometry — and is filed separately as #529.

## Measured impact (A/B via `benches/compare_head_vs_main.sh` vs `origin/main`)

| scenario | Δ | note |
|---|---|---|
| `rust_large/delta_noop` | **−99%** | the genuine win: idle/focus re-requests on an unchanged doc |
| `*/full` (repeated) | **−85…−93%** | real, but partly a bench artifact — editors send `full` once then `full/delta`, not repeated `full` |
| `rust_large/edit_delta` (typing) | **neutral** | +13% in A/B, but an **A/A run (identical binaries) showed this scenario varies ±9–21%** — i.e. it's measurement noise on a high-variance reparse+retokenize path; the added edit-path cost is one FNV-1a hash (~70µs) + one DashMap miss |

The honest framing: a real win on **idle/duplicate re-requests**, no regression on the **typing** path (verified against the noise floor). The user constraint "no performance degradation" is the gate here, and edit→read is neutral.

## Correctness

Two HIGHs found in review and fixed in-branch:
- **Auto-install reload didn't clear the cache** (`/review`): a second `apply_raw_settings` (InstallCoordinator) reloaded queries without invalidating → stale tokens on the post-install refresh. Fixed by moving invalidation into the single `apply_shared_settings` choke point.
- **Settings reload was not atomic with concurrent requests** (`codex`): a request mid-tokenization under the old queries could store stale tokens *after* the reload's clear and poison the post-reload refresh. Fixed with a **settings generation** folded into the cache key: each handler snapshots the generation at its top (before any query/capture-mapping read), the reload bumps it, so tokens from a racing request carry the old-generation key and are invisible to post-reload requests. New test `bump_generation_invalidates_pre_reload_tokens`.

`didChange` deliberately never invalidates (delta needs the previous tokens). Internal-only: identical LSP responses, just without the redundant compute.

## Reviews
`/review`, codex, and pi-review all converged (the two HIGHs above were the only actionable findings; a pi-review TOCTOU flag was a false positive — there is no `.await` between the cache-hit's staleness check and its serve, now documented in-code).

### DIFF
--- a/src/analysis/semantic_cache.rs
+++ b/src/analysis/semantic_cache.rs
@@ -18,13 +18,14 @@ use url::Url;
 #[derive(Clone)]
 pub struct CachedSemanticTokens {
     pub tokens: SemanticTokens,
-    /// FNV-1a hash of the document text these tokens were computed from. Lets a
-    /// repeat request on an *unchanged* document serve the cached tokens instead
-    /// of re-tokenizing (the `result_id` can't do this — it is a fresh global
-    /// counter per response, so it never matches the document's current state).
-    /// Text-only: a config/query reload changes tokenization for the same text,
-    /// so the cache is dropped wholesale on settings apply (see `clear`).
-    pub content_hash: u64,
+    /// Opaque validity key these tokens were computed under (built by
+    /// `CacheCoordinator::token_cache_key`: the FNV-1a hash of the document text
+    /// folded with the settings generation). Lets a repeat request on an
+    /// *unchanged* document under *unchanged* settings serve the cached tokens
+    /// instead of re-tokenizing — the `result_id` can't (it is a fresh global
+    /// counter per response). A text edit OR a config/query reload changes the
+    /// key, so stale tokens stop matching.
+    pub cache_key: u64,
 }
 
 /// Thread-safe semantic token cache.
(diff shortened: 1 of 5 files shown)


atusy:
Applied in f9e7cfa5 — thanks @gemini-code-assist. On a delta cache hit where the cached `result_id` matches `previous_result_id`, we now return the empty delta directly and skip the `previous_tokens` clone + O(N) `calculate_delta`. Safe via the same no-await reasoning as the cache hit; a differing baseline still falls through to the diff. Added a no-op-delta test.

atusy:
Applied in fa66b88b — thanks @gemini-code-assist. `apply_shared_settings` now takes `&LanguageCoordinator`; call sites are unchanged via deref coercion.

atusy:
Addressed in 61cea0fd. The delta short-circuit (matching baseline → empty delta) was already applied in f9e7cfa5 — the re-posted thread was on the shifted line. For the input-consistency point: `capture_mappings` and `supports_multiline` now read alongside the query, before the parse await, in both handlers, so a reload during the await can't split the tokenization inputs. Re-requesting review.

atusy:
Note for the record: the delta short-circuit suggested on `semantic_tokens.rs:642` is already implemented (commit f9e7cfa5) — the `if cached.result_id.as_deref() == Some(previous_result_id.as_str())` fast path right there returns the empty delta and skips the clone + O(N) diff. Functionally identical to the suggestion. All review threads resolved.

PR #635, opened 2026-07-08
fix(completion): preserve incomplete empty results

### LINKED ISSUE #591: completion: {isIncomplete: true, items: []} is classed as "no result", losing the re-query signal (LSP 3.18)
## Found by aggregation audit

- src/lsp/lsp_impl/text_document/completion.rs:81 — virt fan-in emptiness predicate is `!list.items.is_empty()`
- src/lsp/lsp_impl/bridge_context.rs:166-181 — host layer: `items: []` → empty regardless of `isIncomplete`

### Scenario
The higher-priority server returns `{isIncomplete: true, items: []}` ("keep typing, I'll have answers"); it is classed as failed, and a lower-priority server's `isIncomplete: false` list wins the position. The client receives a *complete* list from the wrong server and may filter client-side without re-querying — the higher-priority server's eventual results never surface.

Per LSP 3.18 the `isIncomplete` flag is load-bearing; the emptiness predicate conflates "no items now, re-query" with "nothing".

### Fix direction
Treat `isIncomplete: true` as non-empty in both predicates (or propagate `isIncomplete` into the winning aggregate).

atusy:
## Summary
- treat `CompletionList { isIncomplete: true, items: [] }` as a valid preferred result
- apply the same rule to raw host-layer completion-list values
- add focused regression tests for complete and incomplete empty completion lists

Fixes #591

## Tests
- `cargo test lsp::lsp_impl::text_document::completion::tests`
- `cargo test completion_list_is`
- `cargo fmt --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated completion result detection so **incomplete** suggestion lists are treated as valid results even when they contain zero items.
  * Ensured **completed** suggestion lists with an empty items array are still treated as empty results.
  * Aligned completion-layer handling for consistent “has results” behavior across completion flows.
  * Added unit tests covering both `isIncomplete: true` (non-empty result) and `isIncomplete: false` (empty result).

### DIFF
--- a/src/lsp/lsp_impl/bridge_context.rs
+++ b/src/lsp/lsp_impl/bridge_context.rs
@@ -201,6 +201,14 @@ pub(crate) fn is_empty_layer_value(value: &serde_json::Value) -> bool {
     if let Some(object) = value.as_object() {
         for key in ["items", "signatures", "ranges"] {
             if let Some(list) = object.get(key).and_then(serde_json::Value::as_array) {
+                if key == "items"
+                    && object
+                        .get("isIncomplete")
+                        .and_then(serde_json::Value::as_bool)
+                        .unwrap_or(false)
+                {
+                    return false;
+                }
                 return list.is_empty();
             }
         }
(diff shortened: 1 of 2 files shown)


atusy:
@revise

Issue #582, opened 2026-07-06
bridge: virt-path response transformers silently drop malformed downstream results (8 paths, no log)


atusy:
## Found by error-path audit

The virtual-doc bridge response transformers return `None` on a malformed downstream `result` with **no log line**:

- src/lsp/bridge/protocol/response.rs:35 (and :45/:60/:78 falling through to bare `None` at :93) — `transform_goto_response_to_host`, used by `definition`, `typeDefinition`, `implementation`, `declaration`. A missing `result` field or a `result` that fails to deserialize as `Location | Location[] | LocationLink[]` returns `None` silently.
- src/lsp/bridge/text_document/references.rs:108,124,138 — same shape for `references`.
- src/lsp/bridge/text_document/prepare_rename.rs:115, moniker.rs:101, document_highlight.rs:106 — same pattern via `.ok()?`.

Trigger: a buggy downstream server (third-party binary from user config) returns a JSON-RPC success whose `result` violates the LSP union shape. The client sees "no definition/references/rename" indistinguishable from a legitimate empty result, and nothing is logged — a broken downstream server is undetectable.

This is a confirmed inconsistency, not a style call: sibling transformers in the same module (`transform_formatting_response_to_host`, host.rs `parse_host_formatting_response` / `parse_host_diagnostic_response` / `normalize_host_goto_result` ~line 654) explicitly `log::warn!` and propagate a distinguishable error for the identical failure mode, with doc comments explaining why silent collapse would be wrong. The goto path's JSON-RPC *error* case is already handled (`response_has_jsonrpc_error`, response.rs:32); only the malformed-success case is silent.

### Fix
Add the same warn-on-deserialize-failure treatment to the 8 virt-path transformers.

atusy:
Audited the full 8-path scope against current main. All paths are now covered by merged PRs:

- #650: shared goto transformer used by definition, typeDefinition, implementation, and declaration (4 paths)
- #651: references
- #652: prepareRename
- #653: moniker
- #654: documentHighlight

For each transformer, malformed or missing successful results are now observable via a bridge warning directly or via an io::Error propagated to fan-in logging, while protocol-legal JSON-RPC errors and null results remain legitimate no-result outcomes. The references subset in #651 additionally preserves valid empty arrays and rejects malformed array members.

Verification on current main:
- focused transformer suites: goto 2, references 10, prepareRename 10, moniker 8, documentHighlight 10, all passed
- make check passed
- historical #651 CI passed all checks
[...]

PR #653, opened 2026-07-09
fix(lsp): fail malformed moniker results

atusy:
## Summary
- return `io::Error` for malformed successful `textDocument/moniker` responses
- preserve `Ok(None)` for JSON-RPC errors and legitimate `null` results
- keep valid and empty `Moniker[]` results covered

This addresses the moniker portion of #582. Other transformer families in #582 remain for follow-up PRs.

## Verification
- `cargo test moniker_response --lib`
- `make check`

## Review
- Codex MCP review: no comments
- Fresh Codex MCP review: no comments

### DIFF
--- a/src/lsp/bridge/text_document/moniker.rs
+++ b/src/lsp/bridge/text_document/moniker.rs
@@ -61,7 +61,7 @@ impl LanguageServerPool {
             },
             |response, _ctx| transform_moniker_response_to_host(response),
         )
-        .await
+        .await?
     }
 }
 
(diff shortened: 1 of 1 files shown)


atusy:
Review note on the current bot comments:

- The malformed-success path now returns `io::Error`; preferred fan-in logs that error at warn level as `bridge request failed (...)`, so the downstream protocol/shape bug is observable without duplicating a local warning. I intentionally avoided local `warn!` calls after Codex flagged duplicate warnings on the prepareRename slice.
- The nested-`Result` shape is a valid shared API cleanup candidate, but I am keeping this PR scoped to the moniker bug slice so it stays aligned with the references/prepareRename/documentHighlight PRs and does not mix executor refactoring into this behavioral fix.

Verification already run: `cargo test moniker_response --lib`, `make check`, Codex MCP review, fresh Codex MCP review.

PR #662, opened 2026-07-10
fix(queries): match nvim predicate semantics

### LINKED ISSUE #599: queries: nvim-treesitter semantic divergences — #contains? AND-vs-OR, `; inherits: (lang)` parens, #offset! column base (low)
### 1. `#contains?` uses AND across args and ALL across nodes; nvim uses OR/ANY
src/language/query_predicates.rs:284-291 (`check_contains` uses `Iterator::all`) and :158-159 (`check_match_predicates` uses `nodes().all(...)`). nvim-treesitter's `contains?` returns true on the first matching string arg (OR) and for ANY node of a quantified capture. kakehashi requires every arg to be a substring and every node to satisfy — pinned by test `contains_predicate_multiple_args` (`#contains? @name "foo" "bar"` matches only `foobar`; nvim matches `foo`, `bar`, and `foobar`). Low impact today (multi-arg `#contains?` is rare) but inverted semantics for imported nvim assets.

### 2. `; inherits: c,(cpp)` parenthesized form not stripped — fails the whole language's query load
src/language/query_loader.rs:167-177 (`parse_inherits_directive`) splits on `,` and trims but does not strip the `( )` nvim-treesitter uses for inherit-for-injection-only parents. `(cpp)` becomes a literal language name; `resolve_query_recursive` (:136-141) then errors, failing the **entire** language's query load (not a per-parent drop). No impact on the currently installed languages; a hazard as the corpus expands (PR #557: 7→24 langs).

### 3. `#offset!` row+column applies the column as absolute, not relative (latent)
src/analysis/offset_calculator.rs:107-141 — when `row_offset != 0`, the column offset is applied as `target_line_start + col_offset` instead of nvim's `orig_col + col_offset`. Never fires today (row offsets only authored on column-0 nodes: frontmatter, code_fence_content); would misplace boundaries for a row offset on a non-column-0 node.

atusy:
## Summary
- align `#contains?` with nvim-treesitter OR/ANY semantics
- strip parenthesized `; inherits:` language names such as `(cpp)`
- apply `#offset!` row movement column deltas relative to the original column

Fixes #599

## Verification
- cargo test contains --lib -- --nocapture
- cargo test parse_inherits_directive_strips_parenthesized_language_names --lib -- --nocapture
- cargo test test_apply_offset_to_position_row_offset_preserves_original_column --lib -- --nocapture
- make check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved position/offset calculations to prevent overflow, preserve column behavior across row+column moves, and clamp out-of-range/negative results consistently.
  * Updated `#contains?` matching so it succeeds when **any** node in a capture satisfies the substring rule (including quantified captures).
  * Revised multiple-argument `#contains?` so success occurs if the text contains **any** provided string.
* **Language Support**
  * Improved `inherits` handling: inherited language names wrapped in parentheses (e.g., `(cpp)`) are now normalized correctly.

### DIFF
--- a/src/analysis/offset_calculator.rs
+++ b/src/analysis/offset_calculator.rs
@@ -44,8 +44,8 @@ pub fn calculate_effective_range(
     // Calculate raw effective positions
     let (raw_start, raw_end) = if offset.start_row == 0 && offset.end_row == 0 {
         // Column offsets only - apply directly
-        let start = (byte_range.start as i32 + offset.start_column).max(0) as usize;
-        let end = (byte_range.end as i32 + offset.end_column).max(0) as usize;
+        let start = apply_column_offset(byte_range.start, offset.start_column);
+        let end = apply_column_offset(byte_range.end, offset.end_column);
         (start, end)
     } else {
         // Row offsets require text scanning
@@ -89,7 +89,7 @@ fn apply_offset_to_position(
 ) -> usize {
     if row_offset == 0 {
         // No row offset, just apply column offset
-        return (byte_pos as i32 + col_offset).max(0) as usize;
+        return apply_column_offset(byte_pos, col_offset);
     }
 
     // Start of the line containing byte_pos, found by scanning backward from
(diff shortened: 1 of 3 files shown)


atusy:
@gemini-code-assist review

atusy:
@copilot review

PR #676, opened 2026-07-10
perf(transport): measure stdout head-of-line blocking

### LINKED ISSUE #665: perf(transport): measure and reduce stdout head-of-line blocking from large responses
## What

Measure and, only if the measured ceiling justifies it, reduce server-to-client stdout head-of-line blocking when a large `kakehashi/captures/*` or diagnostic payload starts writing immediately before a small latency-sensitive response such as `textDocument/semanticTokens/*`.

Keep one physical JSON-RPC writer. The problem is response admission/order **before a frame starts**, not a request to interleave frames or add concurrent writes to stdout.

## Evidence

Release measurements on the 38KB / ~5,088-line `example.md` workload:

- semantic tokens alone: p90 approximately **7ms**
- semantic tokens with captures output contention: p90 approximately **17.4ms**
- conservative fixed-timeline replay where old captures `previousResultId`s intentionally miss current lineages and delta falls back to full: semantic p90 approximately **31.2ms**

The likely current ceiling is therefore around **10-24ms at p90**, not the hundreds of milliseconds seen in the original trace. The larger historical delays were primarily compute/cancellation/diagnostic-amplification problems and have separate fixes.

`tower-lsp-server` currently does:

1. handlers complete through `buffer_unordered(max_concurrency)`;
2. responses enter a zero-capacity channel in completion order;
3. one `FramedWrite<stdout>` writes each JSON-RPC frame completely.

Once a multi-megabyte captures frame has started, a semantic response cannot safely overtake it.

## Scope boundary

This issue concerns **kakehashi -> editor stdout**.

It does not replace the downstream bridge's strict FIFO writer (`docs/architecture-decisions/ls-bridge-message-ordering.md`). The downstream writer preserves `didChange -> request` ordering and already skips explicitly cancelled requests that are provably still queued.

## Measure-first work

- Extend the profiling harness to distinguish:
  - handler completion / response-ready time;
  - stdout write start and completion;
  - method, exact frame bytes, and time blocked behind an in-progress frame.
- Establish an A/A noise floor.
- Reproduce at least:
  - semantic-only;
  - semantic + valid captures delta;
  - semantic + captures full/fallback;
  - diagnostics burst.
- Report how often semantic completes before versus after the large frame has already started. Scheduling can help only the former.

## Options, in preferred order

1. **Reduce payload/amplification before the transport**
   - keep captures delta lineages valid;
   - avoid unnecessary captures full re-acquisition;
   - retain diagnostics/captures coalescing and obsolete-work cancellation.
2. **Bounded response-admission scheduler before the single writer**
   - select a ready small/latency-sensitive response before an unstarted large response;
   - preserve a single non-interleaved writer;
   - add aging/bounds so large responses cannot starve.
3. **Captures protocol pagination/chunking**
   - only if client/server protocol changes are acceptable and measurements show full frames remain dominant.

If option 2 requires forking/replacing tower-lsp's transport, include that maintenance coupling in the decision. Do not add method-specific mutable scheduling state unless the measured benefit exceeds the A/A noise floor and payload-side options are insufficient.

## Non-goals / safety constraints

- No multiple tasks writing the same stdout stream.
- No byte-level frame preemption or interleaving.
- Do not reorder stateful notifications or server-to-client requests without proving the dependency boundary.
- Do not conflate this with downstream request/cancel FIFO.
- Preserve JSON-RPC response correlation, cancellation behavior, and shutdown draining.

## Done when

- The response-ready -> write-complete contribution is measured by method and payload size.
- An A/B benchmark demonstrates a statistically meaningful semantic p90 improvement under output contention, **or** the optimization is explicitly deferred with a measured low ceiling.
- Semantic/captures/diagnostic protocol tests remain unchanged in result content.
- Unit, `make check`, and integration/E2E gates pass.

## Related

- #559 — captures layer-tree optimization, also measure-first
- #533 — diagnostic-path instrumentation
- #534 — pull diagnostic `Unchanged`, constrained by cross-document correctness
- `docs/architecture-decisions/ls-bridge-message-ordering.md` — downstream FIFO boundary (not the stdout transport addressed here)

atusy:
## Summary

- add opt-in response-ready, stdout write, flush, frame-size, and censorship metrics behind `KAKEHASHI_STDOUT_METRICS`
- extend the profiling driver with semantic-only, valid captures delta, captures full fallback, and diagnostics burst scenarios plus strict protocol-shape validation
- preserve the single FIFO writer and document why a priority scheduler is deferred under the measured workload

## Measured result

Release measurements used the 5,088-line, 37,940-byte captured-scale Markdown fixture, 20 edit/recompute cycles, and three repetitions per scenario.

- semantic ready to last byte p90: **0.3-0.6 ms** across all scenarios
- semantic ready to flush p90: **0.4-0.7 ms**
- 4,408.4 KiB captures-full frame ready to last byte p90: **14.2 / 15.8 / 14.3 ms**
- semantic responses queued behind an earlier not-yet-started large frame: **0 / 60**
- valid captures delta payload: about **0.3 KiB**, versus **4,408.4 KiB** for full fallback

The interleaved observer A/A check showed no systematic regression: semantic p90 was **9.1 / 9.6 / 9.0 ms** with metrics and **10.2 / 9.0 / 9.4 ms** without metrics. Therefore this PR does not add a scheduler: the observed avoidable FIFO contribution is below the 1 ms decision threshold and there were no reorderable overlaps.

## Validation

- `cargo test --lib`: 2728 passed, 2 ignored
- `make check`: passed, including 10 profiling-driver tests
- release four-scenario matrix and captures-full observer A/A: passed with zero censored or unattributed samples
- `make test_e2e`: 56/57 binaries passed after retry; the remaining `e2e_code_action::concatenated_default_merges_actions_from_both_servers` test added on current `main` independently drops either mock server result and also failed 5/5 focused retries. This branch does not change code-action paths, and metrics-disabled server construction retains the original path.

Fixes #665.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2111,6 +2111,7 @@ dependencies = [
  "bytes",
  "futures-core",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,7 +86,7 @@ tar = "0.4.46"
 tempfile = "3"
 thiserror = "2"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "io-std", "io-util", "time", "process", "fs", "signal"] }
-tokio-util = "0.7"
+tokio-util = { version = "0.7", features = ["rt"] }
 toml = "1"
 tower = { version = "0.5", default-features = false }
 tower-lsp-server = "0.23"
@@ -136,4 +136,3 @@ rstest = "0.26"
 [[bench]]
 name = "semantic_tokens"
 harness = false
-
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 # Variables
 CARGO = cargo
+PYTHON = python3
 TARGET_DIR = target
 RELEASE_DIR = $(TARGET_DIR)/release
 BINARY_NAME = kakehashi
@@ -62,6 +63,7 @@ check:
 	$(CARGO) check
 	$(CARGO) clippy -- -D warnings
 	$(CARGO) fmt --check
+	$(PYTHON) -m unittest benches/profile/test_drive.py
 
 # Format code
 .PHONY: fmt
--- a/benches/profile/README.md
+++ b/benches/profile/README.md
@@ -27,11 +27,14 @@ python3 benches/profile/drive.py \
   --bin ./target/release/kakehashi \
   --file path/to/input.md --requests 20 --edits 1
 
-# Queue a captures delta first, then semantic tokens, to reproduce shared-pool and
-# response-output contention from an already-busy highlighter client.
-python3 benches/profile/drive.py \
-  --bin ./target/release/kakehashi \
-  --file path/to/input.md --requests 20 --edits 1 --concurrent-captures
+# Measure the four stdout head-of-line scenarios from issue #665. The captures
+# modes verify that the response is a real delta or the intended full fallback.
+for scenario in semantic-only captures-delta captures-full diagnostics-burst; do
+  python3 benches/profile/drive.py \
+    --bin ./target/release/kakehashi \
+    --file path/to/input.md --requests 20 --edits 1 \
+    --stdout-metrics --scenario "$scenario"
+done
 
 # Send superseding requests in bursts to measure cancellation pressure.
 python3 benches/profile/drive.py \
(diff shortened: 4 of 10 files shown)


atusy:
Closing without merge. The investigation found no measurable opportunity for a stdout scheduler: semantic transport p90 was 0.3–0.6 ms and schedulable large-frame overlap was 0/60. The measurement branch served its purpose, but retaining the instrumentation and maintenance surface is not justified by the measured benefit.

Issue #747, opened 2026-07-12
bridge: propagate upstream workspace-folder deltas to live connections


atusy:
## Summary

Follow-up to #733: propagate upstream `workspace/didChangeWorkspaceFolders` additions/removals to already-running downstream connections, including incapable-server fallback and shared-instance provenance.

## Why separate

The bridge already adds marker-derived roots to `preferSharedInstance` connections. An upstream folder can have the same URI as a marker-derived root, so blindly forwarding a removal and deleting it from `WorkspaceFolderSet` can remove a root still owned by an open document. Live propagation therefore needs source/provenance tracking rather than a flat URI set.

## Required behavior

- Client-fallback connections that advertise change notifications receive ordered add/remove events.
- Incapable live connections are safely recycled/reinitialized without losing open-document synchronization.
- Shared connections retain a URI while either upstream membership or marker/open-document ownership remains.
- Queue failure does not commit local folder state ahead of the downstream wire.
- Concurrent events preserve upstream order.

## Validation

Use connection-level tests for add/remove, duplicate URI provenance, queue-full rollback, incapable restart, and concurrent ordering.

atusy:
Closing: the core of this landed on main via the #748 lineage. `apply_workspace_folder_change` (src/lsp/bridge/pool.rs) now propagates upstream deltas to capable client-fallback connections in order, recycles initializing/incapable/enqueue-failed connections through the invalidate + `pending_reopen` path (so open-document sync is restored by derive-on-respawn, #928), and commits per-connection folder state only after the notification is queued.

The provenance concern is resolved by scoping rather than tracking: upstream deltas are deliberately **not** forwarded to marker-owned or shared connections at all ("Marker-owned and shared connections derive their folders from marker-root acquisition, not this client snapshot"), so a marker-derived root can never be deleted by an upstream removal — retention holds trivially, with no source-tagged folder set needed. PR #751, which implemented the wider scope (shared propagation + strict register/unregister serialization), is closed with rationale.

The remaining gap — servers that declare the capability only via dynamic `client/registerCapability` are permanently treated as incapable — is tracked in #968 with a lighter design.

PR #750, opened 2026-07-12
feat(lsp): reload config for primary workspace

### LINKED ISSUE #746: workspace: reload primary-root project config after folder changes
## Summary

Follow-up to #733: define and implement primary workspace-root selection and atomic project-config reload when `workspace/didChangeWorkspaceFolders` changes the first active folder.

## Why separate

The bounded #733 lifecycle slice can advertise/consume folder changes and maintain the current folder snapshot. Reloading config is a distinct transaction: it must preserve initialization options and runtime `didChangeConfiguration` layers, change `SettingsManager::root_path`, load the new root `kakehashi.toml`, and apply language/query/downstream settings atomically through the existing reload lock. A naive `load_settings(new_root, None)` would discard non-file layers.

## Required decisions

- First remaining folder becomes primary, preserving client order.
- Define fallback when the last folder is removed (deprecated `rootUri`, process CWD, or no project layer).
- Preserve initializationOptions and runtime configuration layers while replacing only the project-file layer.
- Reparse/refresh and recycle downstream launch configs through the existing settings transaction.

## Validation

Add mixed-layer tests where removing the primary folder switches project TOML without losing initialization/runtime overrides.

atusy:
## Summary

- select the first active workspace folder as the primary project-config root
- fall back to initialize-time `rootUri`, then process CWD, after the last folder is removed
- retain initialization/runtime session overlays separately while replacing only the project-file layer
- publish root, raw settings, and effective settings in one SettingsManager snapshot
- route the reload through the existing language/query/downstream settings transaction

Closes #746.

## Stack

Stacked on #748, which provides the advertised workspace-folder capability and ordered live folder snapshot. Keep this PR draft until #748 merges and this branch is retargeted to `main`.

#714 owns settings-derivation serialization. This PR deliberately does not duplicate that lock. After #714 merges, rebase this branch onto the resulting latest `main` and run the mixed `didChangeConfiguration`/workspace-folder integration checks before marking ready.

## Validation

- `cargo test --lib -q lsp::bridge::workspace::folder_set` (8 passed)
- `cargo test --lib -q did_change_workspace_folders` (3 passed)
- `cargo test --lib -q lsp::settings_manager` (39 passed)
- `cargo fmt --check`

## Safety

Folder removals affect only the upstream folder snapshot. This change does not remove marker-derived folders from live downstream shared connections (#747 owns provenance-safe live propagation). A failed root config expansion retains both the previous root and previous settings.

### DIFF
--- a/src/lsp/bridge/pool.rs
+++ b/src/lsp/bridge/pool.rs
@@ -542,9 +542,7 @@ impl LanguageServerPool {
         &self,
         folders: Option<Vec<tower_lsp_server::ls_types::WorkspaceFolder>>,
     ) {
-        if let Some(folders) = folders {
-            self.workspace_folders.apply_change(folders, &[]);
-        }
+        self.workspace_folders.replace(folders);
     }
 
     /// Get the workspace folders.
--- a/src/lsp/bridge/pool.rs
+++ b/src/lsp/bridge/pool.rs
@@ -372,7 +372,7 @@ pub struct LanguageServerPool {
     ///
     /// Set once via `set_workspace_folders()` after receiving the upstream initialize request.
     /// Passed to downstream servers during LSP handshake.
-    workspace_folders: OnceLock<Option<Vec<tower_lsp_server::ls_types::WorkspaceFolder>>>,
+    workspace_folders: super::WorkspaceFolderSet,
     /// Client capabilities forwarded from upstream client.
     ///
     /// Set once via `set_client_capabilities()` after receiving the upstream initialize request.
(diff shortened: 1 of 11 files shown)


atusy:
Revise status: Codex review confirmed this stacked branch cannot converge before #714 lands. Two concrete races remain because settings derivation currently happens before `SETTINGS_RELOAD_LOCK`: concurrent folder notifications can publish roots out of order, and post-install derivation can overwrite settings composed for a newer root. The correct follow-up is to rebase after #714 and extend its `begin_settings_transaction` across folder mutation/root derivation through publication, then add a concurrent full-handler regression test. This PR remains draft.

One documentation follow-up is also pending: update `downstream-settings-propagation.md`, which still describes project config as startup-only.

atusy:
Superseded. #746 was resolved by #954, which took a deliberately narrower fallback ladder (first folder → `rootUri` → `rootPath`, and no project layer otherwise) rather than this branch's fall-through to the process CWD — a launch-directory `kakehashi.toml` can name parser libraries to `dlopen` and `searchPaths` to auto-install into, so a folder removal must not migrate the session there.

The remaining half of #746 — replacing only the project-file layer while preserving the initialization and runtime layers — is tracked as Defect 1 in #948: those layers survive on main but in already-anchored form, so a relative client path pushed under the old root never rebinds.

The two ordering races noted above (out-of-order root publication across concurrent folder notifications, and post-install derivation overwriting settings composed for a newer root) also remain the property of the settings-transaction work, not of this branch.

Closing rather than rebasing: the branch conflicts with main and both its design decisions have since been made differently.

PR #786, opened 2026-07-12
fix(bridge): preserve workspace marker entries

### LINKED ISSUE #784: bridge: dangling workspace marker symlinks are ignored
## Bug

Bridge workspace-root detection uses `Path::exists()` for configured `workspaceMarkers`. Because `exists()` follows symlinks, a marker directory entry whose target is temporarily or permanently missing is treated as though the marker name is absent.

Workspace markers are directory-entry signals (they may be files or directories; their contents and target are not read). Ignoring a dangling marker makes the bridge fall back to a farther ancestor or the client workspace, so the downstream server can be pooled and initialized under the wrong project root.

## Reproduction

1. Create a project containing `src/main.rs`.
2. Create `.project-root` in the project as a symlink to a missing target.
3. Configure `workspaceMarkers = [".project-root"]`.
4. Resolve the bridge workspace for `src/main.rs`.

## Actual

No marker is found because `Path::exists()` returns false.

## Expected

The `.project-root` directory entry counts as the configured marker regardless of its target state, matching marker semantics and preventing root fallback. Probe the entry without following symlinks and add a focused regression test.

## Scope

This affects bridge workspace-root resolution only and is independent of query asset presence (#779) and language-status recovery (#780).

atusy:
## Summary

- detect workspace marker directory entries without following symlink targets
- keep existing marker priority and ancestor selection unchanged
- add a focused dangling-marker regression

## Why

`Path::exists()` follows symlinks, so a configured marker symlink with a missing target was ignored. The bridge then fell back to a farther/client root and could pool and initialize the downstream server for the wrong project.

## Testing

- `cargo test --lib lsp::bridge::root_markers::tests -- --nocapture`
- `make check`

Closes #784

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace detection now recognizes marker entries even when they are dangling symbolic links.
  * Improved reliability when locating workspace roots during ancestor traversal.

### DIFF
--- a/src/lsp/bridge/root_markers.rs
+++ b/src/lsp/bridge/root_markers.rs
@@ -37,6 +37,14 @@ fn is_valid_marker(marker: &str) -> bool {
         && components.all(|c| matches!(c, std::path::Component::Normal(_)))
 }
 
+/// Whether the configured marker has a directory entry at this ancestor.
+///
+/// Marker contents and symlink targets are irrelevant: the configured name is
+/// the signal, so use non-following metadata to retain dangling symlinks.
+fn marker_entry_exists(path: &Path) -> bool {
+    std::fs::symlink_metadata(path).is_ok()
+}
+
 /// Resolve the marker root for `document_path` following Neovim's
 /// `vim.fs.root` `(string|string[])[]` priority: entries are tried in order,
 /// and each entry is searched up the document's ancestors nearest-first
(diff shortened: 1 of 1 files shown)


atusy:
@coderabbitai review

atusy:
@copilot review

atusy:
@copilot review

atusy:
@copilot review

PR #805, opened 2026-07-12
feat(config): report overwritten output files

atusy:
## Summary

- report `Created` when forced output atomically creates a new path
- report `Overwrote` when forced output replaces an existing regular file
- preserve no-follow symlink protection while making the disposition race-accurate

Closes #801

## Stack

Depends on #804; merge #804 first, then rebase this branch onto current `main`.

## Validation

- `cargo test --bin kakehashi`
- `cargo clippy --bin kakehashi -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved output-file handling to avoid following symbolic links or Windows reparse points.
  - Output writes now clearly indicate whether a file was created or overwritten.
  - Preserved existing output behavior while improving safety across supported platforms.

- **Tests**
  - Added coverage verifying creation, overwriting, and final file contents.

PR #807, opened 2026-07-12
fix(config): publish forced output atomically

### LINKED ISSUE #806: Config output --force can corrupt an existing file after a failed write
## Bug

The `--force --output` path opens an existing regular file with `truncate(true)` and writes the generated content in place. If `write_all` fails after truncation (for example, disk-full or an I/O error), the previous valid configuration/schema is already destroyed and a partial file remains.

This is distinct from #799, which covers failed no-force creation: force mode must preserve an existing destination until the complete replacement is durable enough to publish.

## Expected

Write forced output to a sibling temporary file and atomically replace the destination only after the complete content has been written, while preserving the symlink refusal from #800 and accurate Created/Overwrote reporting from #801.

## Evidence

`write_forced_output` currently selects the overwrite path with `OpenOptions::truncate(true)` before calling `write_all`.

atusy:
## Summary

- write forced config/schema output completely to a sibling temporary file
- atomically publish with no-clobber first to retain exact Created/Overwrote reporting
- preserve an existing destination and its permissions when a write fails
- retain force-mode symlink refusal

Closes #806

## Stack

Depends on #805; merge #805 first, then rebase this branch onto current `main`.

## Validation

- `cargo test --bin kakehashi`
- `cargo clippy --bin kakehashi -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved forced-output writing to avoid partial/corrupted results when a write fails.
  * Prevented forced output from following symbolic links during replacement for safer updates.
  * Ensured forced-output files on Unix keep the expected creation permissions, matching normal file creation behavior.
* **Tests**
  * Added coverage to verify failed writes don’t modify existing output.
  * Added Unix-only checks to confirm permission consistency for forced-output.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -697,77 +697,50 @@ impl WriteDisposition {
 }
 
 fn write_forced_output(path: &std::path::Path, content: &str) -> std::io::Result<WriteDisposition> {
+    write_forced_output_with(path, content, |file, content| {
+        use std::io::Write as _;
+        file.write_all(content.as_bytes())
+    })
+}
+
+fn write_forced_output_with(
+    path: &std::path::Path,
+    content: &str,
+    write: impl FnOnce(&mut std::fs::File, &str) -> std::io::Result<()>,
+) -> std::io::Result<WriteDisposition> {
     if std::fs::symlink_metadata(path).is_ok_and(|metadata| metadata.file_type().is_symlink()) {
         return Err(std::io::Error::new(
             std::io::ErrorKind::InvalidInput,
             format!("refusing to follow output symlink '{}'", path.display()),
         ));
     }
-    #[cfg(unix)]
-    {
-        use std::io::Write as _;
-        use std::os::unix::fs::OpenOptionsExt as _;
-        let create = || {
-            std::fs::OpenOptions::new()
-                .write(true)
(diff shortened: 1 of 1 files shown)


atusy:
@copilot review

atusy:
@coderabbitai review

atusy:
@copilot review

atusy:
@coderabbitai review

atusy:
@copilot review

atusy:
Final gate note: exact HEAD `c6335b7f0d758922161f5c58c98d2b35d044c5f3` has a fresh clean `codex review --base fix/issue-801-output-wording`, all 7 bin tests/clippy/format/diff checks pass, zero unresolved threads, and exact ancestry on Ready PR #805 (`1a50f730e53113e9c3adcf0c7cc22a9a2130e1c2`). Copilot was requested three times but no review was delivered after the repository bot quota warning, so this uses the documented delivery exception. CodeRabbit was explicitly triggered twice but does not review Draft commits. GitHub CI cannot instantiate because the workflow limits pull requests to base `main`; after #805 merges, rebase this PR onto current main and require normal CI before merge.

PR #813, opened 2026-07-12
fix(install): publish metadata cache atomically

### LINKED ISSUE #812: Install metadata cache writes follow symlinks and overwrite their targets
## Bug

`MetadataCache::write` calls `fs::write(cache/parsers.lua, content)`. If that cache entry is a symlink, the write follows it and truncates the symlink target, which may be unrelated to kakehashi.

## Expected

Publish cache content through a sibling temporary file and replace the cache directory entry atomically. A symlink at `parsers.lua` may be replaced, but its target must never be opened or modified. Failed writes must leave the previous cache entry intact.

## Evidence

`src/install/cache.rs` creates the cache directory and then directly invokes `fs::write(self.cache_path(), content)`; duplicate issue/PR searches for metadata-cache symlink handling returned none.

atusy:
## Summary

- write parser metadata cache through a sibling temporary file
- atomically replace the cache entry without following destination symlinks
- preserve previous content after failed writes
- retain ordinary umask-derived creation mode and existing regular-file permissions
- sync file data/metadata before publish and best-effort sync the cache directory afterward

Closes #812

## Validation

- `cargo test --lib install::cache::tests` (7/7)
- `cargo fmt --check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache writes to reduce the risk of corruption or data loss if a write is interrupted.
  * Safely replaces outdated cache links without modifying their targets.
  * Preserves existing cache permissions across updates.
  * Ensures previously valid cached content remains available when a write fails.

### DIFF
--- a/src/install/cache.rs
+++ b/src/install/cache.rs
@@ -63,11 +63,44 @@ impl MetadataCache {
 
     /// Write content to cache.
     pub fn write(&self, content: &str) -> io::Result<()> {
+        self.write_with(content, |file, content| {
+            use std::io::Write as _;
+            file.write_all(content.as_bytes())
+        })
+    }
+
+    fn write_with(
+        &self,
+        content: &str,
+        write: impl FnOnce(&mut fs::File, &str) -> io::Result<()>,
+    ) -> io::Result<()> {
         // Ensure cache directory exists
         fs::create_dir_all(&self.cache_dir)?;
 
-        // Write content
-        fs::write(self.cache_path(), content)?;
+        let cache_path = self.cache_path();
+        let existing_permissions = fs::symlink_metadata(&cache_path)
+            .ok()
+            .filter(|metadata| metadata.file_type().is_file())
+            .map(|metadata| metadata.permissions());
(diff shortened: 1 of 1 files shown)


atusy:
@copilot review

atusy:
@coderabbitai review

atusy:
Final gate note: exact HEAD `3ddfa1e17d4e33cc4e6ef8e2abc4fb77d0a1e565` has a fresh clean `codex review --base main`, exact-head Copilot review, zero unresolved threads, focused 7/7 cache tests plus local clippy/format/diff checks, and authoritative Actions run 29213308776 completed SUCCESS for this exact SHA (Check/Test/Format/Clippy/Audit). CodeRabbit had already reported SUCCESS for this SHA before the explicit re-review command left its duplicate check stale IN_PROGRESS; using the documented duplicate-check exception. Current `origin/main` remains `75a09c99ba628be59c8833eed571924e4b576a54` and is an ancestor; worktree is clean.

PR #815, opened 2026-07-12
fix(install): reject symlink metadata caches

### LINKED ISSUE #814: Install metadata cache reads follow symlinks outside the cache
## Bug

`MetadataCache::read` checks `exists()`, calls `fs::metadata`, and then `read_to_string` on `cache/parsers.lua`. All three follow symlinks. A fresh symlink to an unrelated Lua file outside the cache is therefore accepted as cached nvim-treesitter metadata and can alter language support/install source decisions.

## Expected

Treat a symlink cache entry as absent without reading its target. Only a regular file located at the cache directory entry should be eligible for TTL validation and parsing.

## Scope

This is the read-side sibling of #812, which prevents cache writes from following symlinks. Duplicate issue/PR searches found no read-side cache handling.

atusy:
## Summary

- open metadata cache entries with no-follow semantics
- derive freshness metadata and content from the same opened handle
- reject Windows reparse points explicitly before reading
- preserve missing/stale cache behavior as `None`

Closes #814

## Stack

Depends on #813; merge #813 first, then rebase this branch onto current `main`.

## Validation

- RED observed: symlink cache returned `Some("outside metadata")`
- `cargo test --lib install::cache::tests` (8/8)
- `cargo fmt --check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata cache security by preventing reads from symbolic links and reparse points.
  * Cache reads now return no data when the cached path is not a regular file.
  * Added coverage to verify symlink and reparse-point handling across supported platforms.

### DIFF
--- a/src/install/cache.rs
+++ b/src/install/cache.rs
@@ -43,13 +43,11 @@ impl MetadataCache {
     /// Returns `None` if cache doesn't exist or is stale.
     pub fn read(&self) -> Option<String> {
         let cache_path = self.cache_path();
-
-        if !cache_path.exists() {
+        let mut file = open_cache_file(&cache_path).ok()?;
+        let metadata = file.metadata().ok()?;
+        if !cache_metadata_is_regular(&metadata) {
             return None;
         }
-
-        // Check if cache is fresh based on file modification time
-        let metadata = fs::metadata(&cache_path).ok()?;
         let modified = metadata.modified().ok()?;
         let age = SystemTime::now().duration_since(modified).ok()?;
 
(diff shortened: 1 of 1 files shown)


atusy:
@copilot review

atusy:
@coderabbitai review

atusy:
@copilot review

atusy:
@copilot review

atusy:
Final gate note: exact HEAD `c43fe9dffd0e08a7eae823ee7fd9b5e9b0557772` has a fresh clean `codex review --base fix/issue-812-cache-symlink`, all 8 cache tests/clippy/format/diff checks pass, CodeRabbit reports SUCCESS, zero unresolved threads, and exact ancestry on Ready PR #813 (`3ddfa1e17d4e33cc4e6ef8e2abc4fb77d0a1e565`). Copilot was requested three times but no review was delivered after the repository bot quota warning, so this uses the documented delivery exception. GitHub CI cannot instantiate because the workflow limits pull requests to base `main`; after #813 merges, rebase onto current main and require normal Unix/Windows CI before merge.

PR #825, opened 2026-07-13
fix(install): reject query file symlinks

### LINKED ISSUE #820: install(queries): symlinked highlights file is accepted as complete
## Summary

`query_install_is_complete` uses `std::fs::metadata` on `<data_dir>/queries/<language>/highlights.scm`. Because `metadata` follows symlinks, a symlink to any non-empty regular file outside the query directory is accepted as a complete installed query set.

The non-force install path then returns `AlreadyExists` without repairing the installation. Before returning, it also reads the external target and parses `; inherits:` directives from it, so content outside the managed query directory can unexpectedly trigger dependency downloads.

## Reproduction

On Unix:

```sh
tmp=$(mktemp -d)
mkdir -p "$tmp/data/queries/lua"
printf '; inherits: javascript\n' > "$tmp/outside.scm"
ln -s "$tmp/outside.scm" "$tmp/data/queries/lua/highlights.scm"
XDG_DATA_HOME="$tmp/data-home" kakehashi language install lua --queries-only
```

The exact CLI data-dir wiring varies by invocation/config, but at the library boundary:

```rust
assert!(query_install_is_complete(&queries_dir));
```

currently succeeds for the symlink above. A non-force install treats it as already installed and may parse the outside target's inheritance directive.

## Expected behavior

A managed `highlights.scm` must itself be a regular file. Symlinks (and other non-regular filesystem objects) should be treated as incomplete, without reading their targets. This keeps completeness checks and inherited-query discovery inside the managed query directory.

## Notes

This is distinct from #779, which concerns runtime reporting of a *dangling* query symlink as a missing asset. Here the symlink target exists, and the install subsystem incorrectly accepts it as a completed installation.

atusy:
## Summary

- classify an active query installation as complete only when its managed `highlights.scm` leaf is a regular file, without following symlinks
- keep crash-backup recoverability separate so an interrupted replacement restores the exact owned pre-replacement directory, including the established user-managed symlink state
- cover both contracts with discriminating Unix regression tests

Fixes #820.

## User-visible impact

A non-force query install no longer trusts or reads an external file reached through `<data_dir>/queries/<language>/highlights.scm`. It treats that directory as incomplete and repairs it through the existing staged replacement path.

## Validation

- `cargo test --lib -q symlinked_highlights_file_is_not_a_complete_install`
- `cargo test --lib -q recovery_restores_owned_backup_with_symlinked_highlights`
- `cargo test --lib -q install::queries::tests` (25 passed)
- `cargo clippy -- -D warnings`
- `cargo fmt --check`

`cargo clippy --lib --tests -- -D warnings` is currently blocked on two pre-existing `clippy::single_range_in_vec_init` failures in `src/language/injection/discovery.rs:1576,1586`, outside this diff.

## Review convergence before PR

- local multi-perspective review: no actionable findings
- Codex round 1: clean
- fresh Codex: found crash-recovery regression; fixed in `81fb524f6`
- Codex verification: clean
- fresh Codex after fix: clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation checks to reject query directories where `highlights.scm` is an external symlink.
  * Preserved crash recovery behavior, allowing valid backups to restore the exact previous directory state.

### DIFF
--- a/src/install/queries.rs
+++ b/src/install/queries.rs
@@ -427,6 +427,18 @@ pub fn query_install_is_complete(queries_dir: &Path) -> bool {
         && (queries_dir.join(QUERY_INSTALL_COMPLETE_MARKER).is_file() || metadata.len() > 0)
 }
 
+/// Whether a kakehashi-owned crash backup is safe to restore as the exact
+/// pre-replacement state. Unlike active install completeness, recovery keeps
+/// the existing user-managed symlink contract: the next install can then
+/// classify and repair the restored directory without losing its contents.
+fn query_backup_is_recoverable(queries_dir: &Path) -> bool {
+    let Ok(metadata) = fs::metadata(queries_dir.join("highlights.scm")) else {
+        return false;
+    };
+    metadata.is_file()
+        && (queries_dir.join(QUERY_INSTALL_COMPLETE_MARKER).is_file() || metadata.len() > 0)
+}
+
 /// RAII cleanup for a staging directory: removes it on drop so every error
 /// path (including `?` propagation added later) leaves nothing stranded. On
 /// the success path `replace_query_dir` renames the directory away, making
@@ -810,7 +822,7 @@ fn newest_complete_backup_dir(
         if !generated_backup_matches_language(name, language) {
             continue;
         }
-        if !backup_is_owned(&path) || !query_install_is_complete(&path) {
+        if !backup_is_owned(&path) || !query_backup_is_recoverable(&path) {
             continue;
         }
         let modified = entry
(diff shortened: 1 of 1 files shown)


atusy:
## Final revise status

- Local multi-perspective review converged.
- Codex review converged after fixing crash-recovery finding M1 in `81fb524f6`; the final fresh review returned no comments.
- GitHub CI is fully green: Format, Clippy, Check, Test, and Audit.
- Review threads: 0 unresolved.
- Copilot was requested three times. Two delivered responses explicitly refused review because the requester quota was exhausted; the third accepted request did not deliver a review. This is a documented delivery exception, not an unaddressed finding.
- Gemini likewise reported its daily quota limit and is non-blocking per task instruction.
- Final ancestry audit: HEAD `81fb524f6b07b5fcb8748d49f12cea7d4777a1d8` is pushed exactly and based on current `origin/main` `75a09c99ba628be59c8833eed571924e4b576a54`.

Issue #628, opened 2026-07-08
codeAction/executeCommand: remaining limitations & cleanups (#568 follow-up)


atusy:
Follow-up to #568 (all 8 PRs merged). Grab-bag of deferred codeAction/executeCommand limitations and cleanups surfaced during the PR reviews. None block the merged sequence; each is independently actionable.

## Limitations (real behavior gaps, fail-soft today)
- **Multi-region codeAction.** A request range spanning two injection fences bridges only the **first** intersecting region. `resolve_first_region_overlapping_range` picks one region (preferred semantics); merging actions across multiple overlapped regions is unimplemented. `src/lsp/lsp_impl/bridge_context.rs:1216` (the "#568 stage 7" note predates the actual PR 7, which did cross-*layer*/server concat, not multi-*region*).
- **`workspace/executeCommand` document sync on downstream respawn.** `dispatch_execute_command` forwards without ensuring the target host/virtual doc is open on the (possibly respawned) connection; a respawn between the codeAction and the executeCommand loses doc state → the command computes against nothing (fails soft, no error dialog). Fix = re-resolve the host's injections and `eager_open_virtual_documents` at the `execute_command_impl` layer before dispatch (no command-encoding change, not a connection-variant swap). Flagged by qodo on PR #625.
- **Palette-fired executeCommand.** Only commands surfaced through a bridged action route today; commands the client fires *without* an action context (command palette, keyed off `executeCommandProvider.commands`) need real advertised names + dynamic registration. Deliberate follow-up (`src/lsp/bridge/protocol/command_routing.rs:34`).

## Polish / perf / test
- **`clamp_range_to_region` per-virtual-line offsets** — the intersection treats a region as a rectangle `[start_column, region_end]`, ignoring per-line column offsets (blockquote `> ` prefixes); a range starting on a later injected line before that line's prefix can clamp to out-of-region coords (fail-soft; pre-existing). `bridge_context.rs`. Flagged by Copilot on PR #622.
- **`resolve_all`-per-codeAction fast-path** — when `range.start` is already inside a region, use the cheaper position lookup instead of the whole-document `InjectionResolver::resolve_all`, **preserving** the range clamping. Flagged by qodo on PR #622.
- **Fixed 1s sleep in `tests/e2e_code_action.rs::open_markdown`** — redundant with `code_action_with_retry`'s null-retry loop; drop it (same cleanup applied in PR 8's e2e). Flagged by Copilot on PR #622.
[...]

PR #827, opened 2026-07-13
fix(bridge): reject ambiguous palette command routing

### LINKED ISSUE #823: bridge: palette command collisions route to an arbitrary downstream origin
## Summary

Downstream `executeCommandProvider.commands` names are recorded in the session-wide `CommandOriginRegistry`. The registry currently stores one `ConnectionKey` per raw command name and overwrites it whenever another connection advertises the same name.

As a result, when two live downstream connections (different servers or workspace roots) advertise the same raw palette command, `workspace/executeCommand` is forwarded to whichever handshake registered last. Handshake completion order is nondeterministic, so a user can execute a command against the wrong server or workspace.

Action-encoded commands are unaffected because their command names carry the exact origin and host URI.

## Reachable path

1. Configure two downstream connections that both advertise a common command name (for example, two workspace-root instances of the same server).
2. Both handshakes call `CommandOriginRegistry::register`.
3. The second registration overwrites the first origin.
4. The editor fires the dynamically registered raw command name from its command palette.
5. `dispatch_palette_command` routes it to the overwritten, handshake-order-dependent origin.

The implementation even documents the collision as an accepted limitation, but executing a user-invoked command in an arbitrary workspace is unsafe; ambiguity should fail soft rather than choose an origin.

## Expected behavior

- Track every distinct connection that advertised a raw palette command.
- If exactly one live advertising connection exists, route to it.
- If multiple live connections advertise the same name, log the ambiguity and return `null` without executing the command.
- Re-registering the same command from the same connection must remain idempotent and must not duplicate dynamic registration upstream.
- Action-encoded command routing remains unchanged.

## Overlap audit

Searched all open issues and PRs plus repository issue search for palette command collisions, duplicate command names, and executeCommand origin ambiguity; found no existing coverage.

atusy:
## Summary

A raw palette command name carries no document and no workspace identity, so when
two connections advertise the same name there is nothing in the request to choose
between them. The registry stored one origin per name and let the *last* handshake
overwrite it, so the choice fell to handshake completion order.

- route to the sole LIVE connection whose exact `executeCommandProvider.commands`
  contains the requested name
- refuse — visibly — when several live connections advertise it, instead of using
  nondeterministic handshake order
- record as reconnect candidates only the origins the fallback can actually revive
- revalidate the exact command on the final/reconnected handle so an upgraded
  server cannot receive a command name it has since removed

Closes #823.

## User-visible impact — read this one

**Refusing is now the behaviour whenever several live connections advertise a raw
palette name, and per-root pooling is on by default.** One server rooted at two
workspace folders — an ordinary two-repository session — is enough. Every palette
command that server advertises becomes a no-op for as long as both connections are
live. On `main` those commands ran, against whichever root happened to handshake
last.

That is the trade #823 asks for, and it is the only option that never runs a
command in the wrong workspace: the request genuinely does not say which root was
meant. What this PR adds is that the user is **told**:

```
[kakehashi] command "ruff.applyFix" was not run: several live connections
[ruff@file:///repo-a, ruff@file:///repo-b] advertise it, and the request
carries no workspace context, so kakehashi will not guess which to use
```

delivered as a `window/logMessage` at WARNING, which the editor shows by default.
`log::warn!` alone would not have: the binary installs `env_logger` with no default
directive, so an unset `RUST_LOG` filters at ERROR and the warning is never even
formatted. "Fails soft with a warning" was, for real users, "fails silently".

The structural escape hatch — per-connection encoded palette entries, so each root
gets its own entry and the user disambiguates by picking — is deferred to a stacked
PR; it depends on the two empirical client-behaviour questions in the
`execute-command-routing-token` ADR's Decision–Implementation Gap.

Action-encoded commands are untouched: they embed their exact origin.

## Rebased onto current main (#926 reconciliation)

This branch predated #926 (`route executeCommand by connection, not by document`),
which rewrote most of `workspace/execute_command.rs`. Rebased from a 616-commit lag;
only one commit conflicted.

- **The re-open barrier is kept, and the revalidation reads the handle after it.**
- **#926's PALETTE-FIRST gate** asked `route(..).is_some()` before decoding, to stop
  a downstream id shaped like a routed name from being stripped to its last segment.
  It now asks `is_registered` (a `contains_key`) — the gate runs on every
  `workspace/executeCommand` and only needs the yes/no.

## What local review changed

Ten independent reviewers ran over the rebased branch. The two findings that mattered
most:

**1. The registry vetoed the fallback it existed for.** The no-live reconnect branch
only works for `is_client_fallback()` keys — acquiring with `document_uri: None`
round-trips to that exact key, while a marker-rooted key cannot be rebuilt without a
document. But *every* advertising key was recorded and counted in the "is the
fallback unambiguous?" test, so a marker-rooted key out-voted the one key that would
have worked. Nothing prunes the registry, so the veto was permanent for the session
— a #628 regression on a path #823 never asked to change. The registry now records
only revivable origins; entry presence still answers the gate's separate question.
The same shape appears with config, so origins whose server was deleted from
`languageServers` are filtered before the unambiguity test too.

**2. Every mutation of the actual fix survived the whole suite.** A reviewer ran
mutation testing against all 3257 tests: dropping the ambiguity refusal, letting the
no-live path pick arbitrarily, disabling the palette-first gate, and inverting the
capability predicate were all green. The seam had been cut in the wrong place — a
classifier returned a verdict and the dispatcher decided what to do with it, so the
tests pinned the half that cannot be wrong. `select_palette_route` now returns the
decision itself and the dispatcher only carries it out.

| mutation | before | now |
|---|---|---|
| ambiguity refusal replaced by "pick the first" | SURVIVED | **CAUGHT** |
| no-live path picks an arbitrary recorded origin | SURVIVED | **CAUGHT** |
| palette-first gate disabled | SURVIVED | **CAUGHT** |
| `advertises_execute_command` → `is_none_or` (no caps claims every name) | SURVIVED | **CAUGHT** |
| registry records non-revivable keys (the veto bug) | — | **CAUGHT** |
| reconnect-candidate spawnability filter dropped | — | **CAUGHT** |
| live-candidate spawnability filter dropped | — | **CAUGHT** |
| by-key launch-config check reverted to `None` | — | **CAUGHT** |
| scan-level launch-config filter removed | — | **CAUGHT** |

Also from review: the refusal messages now name the connections (a user cannot fix a
config they cannot identify); `docs/README.md` and the comment beside the upstream
capability declaration no longer document last-advertiser-wins; and the comment
claiming the revalidation protects against a replacement during the re-open wait was
wrong — `server_capabilities` is a `OnceLock`, so that `Arc`'s answer cannot change,
and the `Arc::ptr_eq` check taken with the enqueue is what covers it. The comment now
says so, including a note not to weaken that check on the strength of this one.

## Validation

- `cargo test --lib` — **3273 passed, 0 failed**.
- e2e (CI runs none, so these are local): `e2e_code_action` **45/0**,
  `e2e_host_bridge` **44/0**, `e2e_organize_imports` **27/0**. `e2e_code_action` is
  load-bearing here, not a formality: every unit test builds both sides of the
  launch-config comparison by hand, so they match by construction. The only place the
  real round-trip runs — a config recorded at spawn versus one re-resolved through
  `resolve_with_wildcard` + `merge_bridge_server_configs` at dispatch — is
  `palette_fired_command_routes_to_its_origin_server`. It also holds #926's
  mutation-verified `a_command_does_not_overtake_the_reopen_of_a_respawned_downstream`.
- `cargo clippy --lib -- -D warnings` — clean (the CI invocation).
- `cargo clippy --lib --tests -- -D warnings` — 12 errors, **all pre-existing on
  `origin/main`** (verified by running the same command there) and none in the files
  this PR touches.
- `cargo fmt --check` — clean.
- **Every commit compiles standalone**, verified commit by commit. Four did not
  before: a rename updated one of two call sites and the repair landed four commits
  later, which broke `git bisect` across the span.

## What codex review changed (stage 2)

Two rounds, the second a direct consequence of the first.

**Live candidates were not validated against the current config.** The reconnect
candidates were filtered; the live ones were not — which is the sharper case, since
the settings snapshot is published before `propagate_settings` finishes invalidating
connections. A process spawned from a config that no longer exists can still be
`Ready` in that window, win the route outright, and run a workspace-mutating command
on a superseded server.

**And filtering by name was not enough.** `is_server_spawnable` catches a server
deleted from config, but not one still configured whose `cmd` moved. During
propagation the old and new processes are both `Ready` and both advertise the
command, so the count is two and the command is refused as ambiguous — although
exactly one is valid. Cardinality is the whole decision here, so anything the
acquisition would reject has to be rejected *before* the count: the scan now applies
the same launch-config comparison the by-key acquisition applies. The reconnect
branch's own spawnability re-check goes with it (the chosen key passed the identical
predicate on the same borrowed snapshot), and the config resolution moves up and is
bound fail-closed.

Codex also confirmed clean: `try_send` on the window channel is non-blocking and the
ordering-sensitive traffic uses a different channel; nothing conflates "empty
reconnect list" with "unknown name".

This round exposed a defect in the tests that came with it — they drove dispatch with
`WorkspaceSettings::default()`, in which no server is configured at all, so once both
sets were filtered the ambiguity test passed by refusing for the wrong reason. They
now configure the servers they seed.

## Refuted during review, recorded so it is not re-derived

- **"`advertises_execute_command` ignores dynamic registrations."** The registry is
  seeded only from static `executeCommandProvider.commands`, and
  `DynamicCapabilityRegistry` matches by method name without parsing
  `ExecuteCommandRegistrationOptions.commands`. A dynamically-registered name can
  never be in the registry to reach this path — and could not on `main` either.
- **"Prune the registry when a connection dies."** Nothing un-registers the name with
  the editor, so the palette can fire it all session; the gate must keep recognising
  it or a downstream id shaped like a routing token gets decoded and misrouted.
- **"`Vec::contains` in `register` is O(n)."** `n` is the number of distinct
  connections that ever advertised that one name, and `register` runs once per
  handshake, never on a request path.
- **"Return a `ResponseError` instead of null."** The spec permits it, but every
  failure branch in this file answers null; making ambiguity the sole error-returning
  one would be the inconsistency. Worth revisiting for the whole file, not here.

## Summary by CodeRabbit

* **New Features**
  * Improved command routing across connected and reconnectable language servers.
  * Reconnects commands only when a single valid destination is available.
  * Validates advertised commands and launch configurations before execution.

* **Bug Fixes**
  * Prevents accidentally invoking commands with duplicate or ambiguous identifiers.
  * Avoids routing commands to stale, initializing, or incompatible connections.
  * Reports rejected or ambiguous command requests directly in the editor.

* **Documentation**
  * Clarified duplicate command handling and configuration-related routing behavior.


atusy:
@coderabbitai review

atusy:
## Revise handoff (exact HEAD `2979aa7ccba2bf66d19baf53d4dae4a75c81b568`)

All available gates are clean:

- exact base: `origin/main` = `75a09c99ba628be59c8833eed571924e4b576a54`
- local multi-perspective review: converged after fixing the Ready/registry publication window and documentation drift
- independent Stage-1 reviewer: no actionable findings
- independent Codex: fixed exact-command revalidation after reconnect/replacement; fresh session returned `No comments to provide`
- CodeRabbit: completed with zero inline findings
- GitHub Actions: Check, Format, Clippy, Audit, and Test all passed
- unresolved review threads: 0
- Gemini: quota-limited (explicitly waived by the user)

Copilot is the only unmet required revise stage. It returned the same quota-limit response twice for this exact SHA; a third review request was accepted but has not delivered a review. Keep this PR draft until Copilot quota resets. Then re-request Copilot, fix/reply/resolve any findings, confirm exact-base ancestry and all checks/threads again, and mark ready only after a clean Copilot round.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
@greptileai review

atusy:
### Known gap, fixed in the stacked follow-up

A fresh independent codex review found one MED that this PR does **not** close, and I am deliberately leaving it here rather than growing this branch again.

**Scenario.** Client-fallback servers A and B both once advertised `source.fixAll`, so both are recorded as reconnect candidates. B restarts after an upgrade and no longer advertises the name; A then crashes. The live scan finds no advertiser (B is live but does not advertise it), while `reconnectable` is still `[A, B]` — so the command is refused as ambiguous instead of reconnecting A. The exact-name revalidation happens after selection and cannot repair it.

It fails **soft** (a refusal, never a misroute), and it needs a specific multi-step history, which is why it is not a blocker.

**Why not here.** The fix is to reconcile each origin's recorded command set on every successful handshake — a connection that comes back without a name should stop being a candidate for it — while keeping the command-NAME registration separate, since the editor still holds that. That is the same machinery the stacked follow-up introduces (per-connection, reversible registration), so doing it here would mean writing it twice.

Tracked as part of the stacked PR; this comment is the record so the gap is not rediscovered as new.

atusy:
The deferred gap is now closed in the stacked follow-up: #950 (`register` reconciles a connection's recorded candidates at handshake, so a server that comes back without a command stops being a candidate for it).

PR #837, opened 2026-07-13
fix(install): confine query uninstall tombstones

atusy:
## Summary

- reject pre-existing tombstone symlinks, reparse points, special files, and multiply linked regular leaves before uninstall
- publish `ok` through a private same-parent inode so no pre-existing inode is truncated, including a hard-link alias raced after validation
- preserve symlinked `data_dir` / `queries` parents while constraining only the managed tombstone leaf
- verify the published inode against a no-follow final reopen; on Windows, pin parent/stage handles and replace the basename with `FileRenameInfo`
- preserve Unix existing modes/new-file umask behavior and normalize Windows tempfile attributes

Closes #822

## Security boundary

This PR prevents this call from opening, following, truncating, or writing an outside victim through the managed tombstone leaf. Whole-transaction confinement against a concurrently retargeted writable parent is separately tracked by #830.

## Follow-ups found during review

- #829: make uninstall-intent directory publication power-loss durable before destructive removal
- #830: bind the full query transaction to one parent-directory identity
- #832: recover crash-stranded named stages only with verifiable ownership
- #836: preserve full Windows owner/DACL metadata through handle-bound replacement

## Validation

- strict RED: symlink leaf overwrote victim (`keep me` became `ok`)
- strict RED: hard-link alias raced after validation was truncated
- strict RED: substituted private stage was accepted
- strict RED: existing Unix `0600` mode became `0644`
- `cargo test --lib -q tombstone_` (8/8)
- `cargo test --lib -q remove_query_install_` (5/5)
- `make check`
- full library suite: 2903 passed, 2 ignored; 5 unrelated current-main failures (missing rust fixture and four closed-host bridge lifecycle tests)
- Stage 1 multi-perspective review: clean at `c029c61d4`
- Codex continued review: clean after findings fixed/refuted
- Codex fresh review: first-response `No comments`

## Platform note

The Windows release target is not installed locally. Windows code was fact-checked against the `windows-sys` API and reviewed independently; cfg-gated regressions cover parent symlinks, destination reparse replacement, and clearing `FILE_ATTRIBUTE_TEMPORARY` when run on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved uninstall reliability by making tombstone updates atomic, reducing the risk of incomplete or corrupted uninstall state.
  * Added safeguards against symlinks, hard links, file replacement races, and other filesystem manipulation scenarios.
  * Preserved appropriate file permissions and cleaned up temporary files when updates fail.

* **Platform Support**
[...]

PR #840, opened 2026-07-13
fix(install): persist query uninstall intent

### LINKED ISSUE #829: install(queries): fsync uninstall intent before destructive removal
## Summary

Query uninstall writes `.LANG.uninstalled` and then removes the installed query directory/backups, but it does not establish crash durability for the tombstone directory entry before destructive removal. A power loss can therefore leave the deletion durable while the tombstone creation/replacement is not, allowing startup recovery to restore a backup contrary to the completed uninstall intent.

This predates #822: the original `File::create` + `write_all` path synchronized neither file contents nor the parent directory. #822 improves content staging and prevents leaf-link victim writes, but same-parent atomic replacement still needs an explicit durability contract.

## Required design

- flush tombstone contents before publication;
- make publication of the `queries/.LANG.uninstalled` directory entry durable before deleting the canonical query directory or backups;
- define Unix directory-fsync behavior and a Windows equivalent (for example replace/move write-through semantics), including filesystems/platforms that reject directory flushes;
- preserve symlinked `data_dir` / `queries` directory support;
- add a fault-injection/order test proving destructive removal cannot begin before durable intent publication succeeds.

## Scope boundary

This is separate from #822’s security invariant: #822 prevents this call from following/truncating a pre-existing symlink or outside hard-link victim. This issue governs power-loss persistence ordering.

atusy:
## Summary

- flush the published tombstone parent directory before query removal on Unix-like platforms
- issue Windows `FileRenameInfo` through the same `FILE_FLAG_WRITE_THROUGH` stage handle so rename metadata is flushed before return
- propagate durability failures before removing canonical queries, owned backups, or backup ownership evidence
- add deterministic sync-failure and destructive-ordering regressions

Closes #829

## Stack

Depends on #837. This PR is intentionally based on `fix/issue-822-query-tombstone` at exact parent head `c029c61d4`; after #837 merges, rebase this branch onto current `main` and rerun revise/CI.

## Validation

- strict RED: injected parent-sync callback was not called
- `cargo test --lib -q tombstone_` (9/9)
- `cargo test --lib -q remove_query_install_` (5/5)
- `cargo test --lib -q uninstall_sync_failure_preserves_queries_and_backups`
- `make check`
- Stage-1 durability review: clean at `fe1a45b14`
- Codex continued review: `No comments`
- Codex fresh review: first-response `No comments`

## Durability contract

Unix ordering is staged-file sync → atomic publish → published identity verification → parent-directory `sync_all` → destructive removal. Unsupported directory sync errors fail the uninstall safely.

Windows reopens the already-created stage handle with `FILE_FLAG_WRITE_THROUGH`; Microsoft documents that metadata changes, including renames, issued through a write-through file handle are flushed rather than lazily persisted. The same handle performs `SetFileInformationByHandle(FileRenameInfo)`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when uninstalling queries and their backups.
  * Added safeguards to ensure uninstall records are durably saved before removal completes.
  * If storage synchronization fails, the system now reports the failure and preserves managed queries and backups.
  * Improved consistency of uninstall behavior across supported platforms, including Windows.

### DIFF
--- a/src/install/queries.rs
+++ b/src/install/queries.rs
@@ -757,13 +757,14 @@ fn write_uninstall_tombstone(
     queries_parent: &Path,
     language: &str,
 ) -> Result<(), QueryInstallError> {
-    write_uninstall_tombstone_with_before_publish(queries_parent, language, |_, _| {})
+    write_uninstall_tombstone_with_before_publish(queries_parent, language, |_, _| {}, |_| Ok(()))
 }
 
 fn write_uninstall_tombstone_with_before_publish(
     queries_parent: &Path,
     language: &str,
     before_publish: impl FnOnce(&Path, &Path),
+    _sync_parent: impl FnOnce(&Path) -> std::io::Result<()>,
 ) -> Result<(), QueryInstallError> {
     validate_safe_language_name(language)?;
     let path = uninstall_tombstone_path(queries_parent, language);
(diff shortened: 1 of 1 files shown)


atusy:
@coderabbitai review

PR #846, opened 2026-07-13
fix(install): reject managed parser symlinks

### LINKED ISSUE #842: managed parser symlinks are accepted as installed parsers
## Summary

The managed parser leaf `<data-dir>/parser/<language>.<DLL_EXTENSION>` is classified through `Path::exists()` / `Path::is_file()`, both of which follow symbolic links. A symlink to any external regular file is therefore accepted as an installed parser.

This is distinct from #828, which fixes directory-versus-file confusion in CLI status, and from #835, which covers the query-language directory leaf.

## User impact

- `language status` reports `✓ parser` for a managed symlink whose target is outside the data directory.
- `parser_file_exists`, used by runtime/auto-install paths, suppresses repair and returns the external target path through the managed link.
- Non-force parser installation returns `AlreadyExists` instead of repairing the managed entry.
- `language uninstall` removes only the symlink but previously advertised the external target as an installed managed parser.

## Evidence on current code / active #828 stack

- `src/install/parser.rs::parser_file_exists` uses `Path::exists()`.
- `install_parser` uses `parser_file.exists() && !force` for its non-force gate.
- `src/bin/main.rs::find_parser_file` uses `Path::is_file()` in #828; that follows a symlink target.
- Parser discovery similarly treats the resolved target as a file.

## Expected behavior

The managed parser leaf itself must be a regular file, not a symbolic link or Windows reparse point. Status/runtime discovery must ignore linked leaves, non-force install must proceed to safely replace/repair them, and uninstall discovery must not claim external targets as managed installations.

atusy:
Fixes #842

## Summary
- reject symlink/reparse-point parser leaves in install and CLI discovery
- repair managed links safely while preserving ordinary directories and special files
- reject managed links during implicit runtime parser resolution while preserving explicit parser overrides
- cover status, explicit/bulk uninstall, runtime lookup, and platform-specific replacement behavior

## Validation
- `cargo test --lib parser_file_exists`
- `cargo test --lib implicit_library_resolution_rejects_managed_parser_symlinks`
- `cargo test --lib test_resolve_library_path`
- focused CLI status/uninstall tests
- `make check`
- fresh exact-parent Codex review: no remaining actionable correctness findings

## Stack
Depends on #831.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1005,6 +1005,7 @@ dependencies = [
  "ulid",
  "ureq",
  "url",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,9 @@ percent-encoding = "2"
 rayon = "1"
 futures = "0.3.31"
 
+[target.'cfg(windows)'.dependencies]
+windows-sys = { version = "0.61", features = ["Win32_Storage_FileSystem"] }
+
 # Unix-only dependencies (signal handling; "fs" for FD_CLOEXEC in tests)
 [target.'cfg(unix)'.dependencies]
 nix = { version = "0.31", features = ["signal", "process", "fs"] }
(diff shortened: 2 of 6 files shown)


atusy:
@gemini-code-assist /review

PR #870, opened 2026-07-19
feat(worker): cut tree readers over atomically

atusy:
## Summary

- add guarded legacy, shadow, and authoritative modes and route all tree-derived readers through one fixed resident worker
- stop parent reparsing after authoritative didChange while preserving virtual-document and diagnostic workflows from worker-owned facts
- keep strict incarnation, content-version, configuration-generation, and worker-generation fencing with opaque node identities
- use binary worker frames and lazy worker-owned semantic, injection, discovery, and capture facts

## Stack

- Depends on #869
- Implements the guarded all-reader cutover stage from ADR #862
- Remains draft because later stacked stages still remove parent-side parser ownership and finish recovery and performance gates

## Measurement

Release benchmark, 40 iterations after 10 warmups:

| workload | legacy | authoritative 4 threads |
| --- | ---: | ---: |
| markdown full sync | 0.305 ms | 0.339 ms |
| large Rust edit delta median | 29.568 ms | 64.686 ms |
| markdown edit delta | 7.009 ms | 11.746 ms |

Large Rust authoritative cache hits were about 0.725 ms at p25, but cache misses and edits remain slower. This stage therefore does not pass the general performance gate. Detailed results and next targets are recorded in the Stage 7 measurement note.

## Validation

- cargo test --lib: 3111 passed, 2 ignored
- cargo fmt --check
- git diff --check

## Rollback

Set KAKEHASHI_TREE_WORKER_MODE=legacy to select the complete legacy path. The cutover is atomic; mixed reader ownership is not a supported production mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -999,6 +999,7 @@ name = "kakehashi"
 version = "0.8.0"
 dependencies = [
  "arc-swap",
+ "bincode",
  "clap",
  "dashmap",
  "dirs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ opt-level = 3
 
 [dependencies]
 arc-swap = "1"
+bincode = "1.3"
 clap = { version = "4", features = ["derive"] }
 dashmap = "6"
 dirs = "6"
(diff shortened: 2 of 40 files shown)


atusy:

PR #871, opened 2026-07-20
perf(worker): consolidate authoritative ownership

atusy:
## Summary

- move authoritative didOpen parsing and install recovery into the resident worker while keeping the parent document tree-less
- preserve worker document replicas across true incremental edits instead of accidentally full-syncing every advancing version
- reuse exact current-version semantic facts, skip empty injection derivation, and acknowledge authoritative edits without shadow snapshots
- expose semantic queue, compute, and cache telemetry and record paired Stage 8 measurements

## Stack

- Depends on #870
- Advances the ownership and performance stage from ADR #862
- Remains draft because authoritative grammar/query metadata loading still starts in the parent and the performance, failure, protocol, compatibility, and memory gates are not complete

## Measurement

Release benchmark, same Stage 8 binary, 40 iterations after 10 warmups:

| workload | legacy | authoritative 4 threads |
| --- | ---: | ---: |
| markdown full | 0.270 ms | 0.303 ms |
| large Rust edit delta | 25.326 ms | 39.016 ms |
| markdown edit delta | 6.853 ms | 7.542 ms |
| markdown cold open + first token | 21.832 ms | 27.590 ms |
| large Rust cold open + first token | 527.047 ms | 554.410 ms |

Stage 8 improves the Stage 7 large Rust edit median from 64.686 ms to 39.016 ms after fixing the replica identity check, but it still fails the paired legacy non-regression gate. Exact-version worker cache hits reach 0.635 ms at p25. Telemetry shows semantic computation around 19-25 ms while uncontended queue wait is only 6-12 microseconds, so the evidence does not justify more workers.

## Validation

- cargo test --lib: 3115 passed, 2 ignored
- cargo fmt --check
- cargo check --lib
- git diff --check

## Rollback

Select legacy mode to restore the complete in-process path. The authoritative and shadow paths retain strict document incarnation, content version, configuration generation, and worker generation fences.

### DIFF
--- a/benches/semantic_tokens.rs
+++ b/benches/semantic_tokens.rs
@@ -59,11 +59,16 @@ struct Server {
 impl Server {
     /// Spawn `bin`, run the LSP handshake, and return a ready server.
     fn start(bin: &str, data_dir: &str) -> Server {
+        let stderr = if std::env::var_os("KAKEHASHI_BENCH_SERVER_STDERR").is_some() {
+            Stdio::inherit()
+        } else {
+            Stdio::null()
+        };
         let mut child = Command::new(bin)
             .env("KAKEHASHI_DATA_DIR", data_dir)
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
-            .stderr(Stdio::null())
+            .stderr(stderr)
             .spawn()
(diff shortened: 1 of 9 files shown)


atusy:

PR #872, opened 2026-07-20
feat(worker): move native loading behind worker boundary

atusy:
## Summary

- resolve immutable parser artifacts and raw query sources in the parent without loading a Tree-sitter Language
- discover search-path parser identities as data and load/compile the complete language catalog only inside the resident worker
- treat worker assets as language availability while keeping the parent parser registry and compiled query store empty in authoritative mode
- preserve tolerant query semantics in the worker and cache raw-input identity separately from filtered compiled source
- make authoritative semantic full, delta, and range bypass parent Tree/query gates and strengthen E2E assertions against empty-token false positives

## Stack

- Depends on #871
- Completes the guarded native grammar/query loading ownership boundary from ADR #862
- Remains draft because cold-open and large serial-miss performance, protocol cancellation/fairness, memory, compatibility, and final failure gates remain open

## Measurement

Release benchmark, same Stage 9 binary, 40 iterations after 10 warmups:

| workload | legacy | authoritative 4 threads | change |
| --- | ---: | ---: | ---: |
| markdown full | 0.266 ms | 0.237 ms | -11% |
| markdown edit delta | 7.604 ms | 7.150 ms | -6% |
| large Rust edit delta | 30.779 ms | 39.265 ms | +28% |
| markdown cold open + first token | 22.675 ms | 33.679 ms | +49% |
| large Rust cold open + first token | 548.209 ms | 597.798 ms | +9% |

This is the first paired end-to-end evidence that one fixed worker improves injection-heavy steady-state workloads. It still fails the general non-regression gate on Rust misses and cold open. An intermediate empty-token response measured around 0.7 ms; the E2E contract now requires non-empty full and range tokens, and those invalid numbers are excluded.

Raw-versus-compiled query identity was also material: recompiling tolerant raw queries on every request regressed Rust edit to 101.229 ms and Markdown edit to 17.398 ms. Caching the raw input restored the final values above.

## Validation

- cargo test --lib: 3120 passed, 2 ignored
- cargo fmt --check
- cargo check --lib
- git diff --check

## Rollback

Select legacy mode to restore the complete in-process parser and query path. Authoritative mode keeps the parent document tree-less and no longer dynamically loads native grammars or compiles Tree-sitter queries in the parent.

### DIFF
--- a/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage9-measurement.md
+++ b/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage9-measurement.md
@@ -1,0 +1,97 @@
+# Single Tree Worker Stage 9 Native-Loading Measurement
+
+## Scope
+
+Stage 9 moves dynamic grammar loading and tolerant query compilation out of the
+authoritative parent. The parent resolves immutable parser artifacts and raw
+query text as data; the worker alone loads `tree_sitter::Language`, compiles
+queries, owns parser registries, and serves tree-dependent readers.
+
+Search-path parser filenames populate the worker catalog without `dlopen` in
+the parent. This preserves injected-language discovery when the parent registry
+is intentionally empty.
+
+## Method
+
+The same release Stage 9 binary was run in both modes on macOS:
+
+```text
+iterations = 40
+warmup = 10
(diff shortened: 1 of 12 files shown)


atusy:

PR #873, opened 2026-07-20
Stage 10: cancel obsolete worker reads

atusy:
## Summary

- add a versioned idempotent cancellation frame for queued and running reader work
- propagate LSP cancellation, supersession, didClose, and handler drop across the process boundary
- keep cancellation readable under bounded load and prevent one document backlog from reserving every permit
- add a running-cancellation benchmark and Stage 10 ADR measurement
- extend bounded cold-readiness waiting so healthy workers do not return null under contended startup

## Measurement

With 40 measured iterations and 10 warmups, an xlarge Rust burst with four already-running requests explicitly cancelled improved final usable response latency from 302.323 ms on Stage 9 to 199.860 ms on Stage 10 (-33.9%). Ordinary workloads remain mixed; large Rust edit was still +23.1%, so the ADR stays proposed.

## Validation

- `cargo test --lib --quiet` (3123 passed, 2 ignored)
- `cargo test --test tree_worker_process --features e2e` (4 passed)
- `cargo test --test e2e_tree_worker_shadow --features e2e --quiet` (33 passed)
- release paired benchmark, 40 iterations / 10 warmups

### DIFF
--- a/benches/semantic_tokens.rs
+++ b/benches/semantic_tokens.rs
@@ -125,6 +125,13 @@ impl Server {
         )
     }
 
+    fn send_semantic_full(&mut self, uri: &str) -> i64 {
+        self.send_request(
+            "textDocument/semanticTokens/full",
+            json!({ "textDocument": { "uri": uri } }),
+        )
+    }
+
     /// `semanticTokens/full/delta` against `previous_result_id`.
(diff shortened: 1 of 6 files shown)


atusy:

PR #874, opened 2026-07-20
Stage 11: prove hard-deadline worker recovery

atusy:
## Summary

- add a deterministic non-cooperative worker hang fixture
- allow E2E-only request deadlines scoped to the implicated URI
- prove timeout kill, native-evidenced classification, session quarantine, restart, and full resync of an unrelated open document
- record the Stage 11 recovery measurement and remaining hazard/control-lane gates

## Measurement

With a 250 ms injected deadline, generation 2 full-resynced one already-open Lua document (23 bytes) in 856 ms on the debug E2E build after the Rust request hung and was quarantined.

## Validation

- `cargo test --test e2e_tree_worker_shadow --features e2e --quiet` (34 passed)
- `cargo test --test tree_worker_process --features e2e --quiet` (4 passed)
- `cargo test --lib request_timeout_override_is_positive_milliseconds --quiet`

### DIFF
--- a/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage11-measurement.md
+++ b/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage11-measurement.md
@@ -1,0 +1,81 @@
+# Single Tree Worker Stage 11 Hard-Deadline Measurement
+
+## Scope
+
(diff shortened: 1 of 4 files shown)


atusy:

PR #875, opened 2026-07-20
perf(worker): preserve cross-document fairness

atusy:
## Summary

- keep nested semantic fan-out for the single-document fast path
- suppress injection fan-out when another document is already runnable
- prove a user-blocking node request completes before injection-heavy work
- record paired Stage 12 measurements and remaining dynamic-fairness gates

## Measurement

Ten isolated debug E2E runs using the same Stage 12 binary:

| Mode | Foreground median | Heavy completion median |
| --- | ---: | ---: |
| fairness protection | 385.4 ms | 1.535 s |
| nested fan-out forced | 424.9 ms | 1.594 s |
| change | -9.3% | -3.7% |

This is admission-time protection. Late-arrival chunking, priority classes, and the one-thread cooperative guarantee remain open.

## Validation

- `cargo test --lib` (3126 passed, 2 ignored)
- `cargo test --test tree_worker_process` (2 passed)
- `cargo test --features e2e --test e2e_tree_worker_shadow -- --nocapture` (35 passed)

### DIFF
--- a/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage12-measurement.md
+++ b/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage12-measurement.md
@@ -1,0 +1,81 @@
+# Single Tree Worker Stage 12 Cross-Document Fairness Measurement
+
+## Scope
+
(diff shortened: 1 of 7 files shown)


atusy:

PR #876, opened 2026-07-20
perf(worker): yield nested fanout at chunk boundaries

atusy:
## Summary

- split semantic injection misses into compute-width chunks
- recheck cross-document competition before every chunk
- resume nested parallelism after competitors leave
- prove late-arrival yielding with a causal E2E barrier
- record no-competitor A/B measurements and remaining scheduler gates

## Measurement

Authoritative worker, 40 requests after 10 warmups, Stage 12 vs Stage 13:

| Scenario | Stage 12 | Stage 13 | Change |
| --- | ---: | ---: | ---: |
| Markdown 60 injections/full | 2.545 ms | 2.537 ms | -0.3% |
| Markdown 150 injections/full | 5.650 ms | 5.640 ms | -0.2% |

The late-arrival E2E test causally registers B as runnable while A's first nested chunk occupies the pool, then proves A yields at the next chunk boundary.

## Remaining gates

- per-document max-min chunk admission across more than two documents
- explicit user-blocking/speculative priority classes
- cooperative chunking for other long tree walks
- one-thread priority/fairness behavior

## Validation

- `cargo test --lib` (3127 passed, 2 ignored)
- `cargo test --features e2e --test tree_worker_process` (4 passed)
- `cargo test --features e2e --test e2e_tree_worker_shadow -- --nocapture` (36 passed)

### DIFF
--- a/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage13-measurement.md
+++ b/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage13-measurement.md
@@ -1,0 +1,91 @@
+# Single Tree Worker Stage 13 Dynamic Fairness Measurement
+
+## Scope
+
+Stage 13 turns Stage 12's admission-time semantic policy into a dynamically
+rechecked policy. Injection cache misses are divided into chunks no larger than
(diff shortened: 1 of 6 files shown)


atusy:

PR #877, opened 2026-07-20
perf(worker): prioritize foreground work at chunk boundaries

atusy:
## Summary

- classify worker requests as foreground or background at reader acceptance
- yield semantic injection fanout once when another document has foreground work
- keep background-only competition sequential and restore parallelism when contention clears
- preserve a constant-time single-document fast path
- prove P=1 cooperative order and deterministic late-arrival priority behavior

## Measurement

Authoritative worker, paired policy fixture with five runs per condition:

| Compute threads | Policy | Foreground | Background |
| ---: | --- | ---: | ---: |
| 4 | Stage 14 priority | 1.343 s | 3.602 s |
| 4 | dynamic policy disabled | 1.839 s | 2.305 s |
| 1 | Stage 14 priority | 4.872 s | 5.332 s |
| 1 | dynamic policy disabled | 4.873 s | 5.125 s |

With four threads, foreground latency improved 27.0% at the intentional cost of 56.3% later background completion. With one compute thread, foreground latency was unchanged; the one-thread path is a correctness fallback, not a performance claim.

Order-balanced no-competitor benchmark, 100 requests after 20 warmups per ordering:

| Scenario | Stage 13 | Stage 14 | Change |
| --- | ---: | ---: | ---: |
| Markdown 60 injections/full | 1.223 ms | 1.224 ms | +0.1% |
| Markdown 150 injections/full | 2.798 ms | 2.711 ms | -3.1% |

## Remaining gates

- memory-pressure-aware admission
- weighted fairness across more than two documents
- cooperative scheduling for other long tree walks
- compatibility gates and legacy-path removal

## Validation

- cargo fmt --all -- --check
- cargo check --lib --tests --features e2e
- cargo test --lib (3130 passed, 2 ignored)
- cargo test --features e2e --test tree_worker_process (4 passed)
- cargo test --features e2e --test e2e_tree_worker_shadow (37 passed)

### DIFF
--- a/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage14-measurement.md
+++ b/docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage14-measurement.md
@@ -1,0 +1,117 @@
+# Single Tree Worker Stage 14 Priority Measurement
+
+## Scope
+
+Stage 14 distinguishes foreground worker requests from bulk derivations.
+Semantic-token injection fan-out now chooses one of three policies at every
+compute-width chunk boundary:
+
+* no competing document: continue parallel fan-out;
+* a background-only document competes: process the chunk sequentially; or
+* a foreground document competes: yield once to the Rayon pool, then process
+  the chunk according to the newly observed policy.
+
+Document writes, node accessors, captures, selection ranges, native bindings,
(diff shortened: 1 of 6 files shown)


atusy:

PR #878, opened 2026-07-20
bench(worker): measure aggregate process memory

atusy:
## Summary

- add a process-tree memory collector with alternating legacy/authoritative order
- sample aggregate RSS and macOS shared-aware physical footprint
- hold initialized workloads briefly for steady-state observation
- retain six-batch release results for 32 KiB and 1.07 MiB Markdown documents
- document why the retained-text limit is not a total-memory bound

## Measurement

Shared-aware physical-footprint medians:

| Workload | Legacy | Authoritative | Change |
| --- | ---: | ---: | ---: |
| 32 KiB Markdown, stable | 16.77 MiB | 23.42 MiB | +6.65 MiB (+39.6%) |
| 1.07 MiB Markdown, stable | 190.67 MiB | 224.06 MiB | +33.38 MiB (+17.5%) |
| 1.07 MiB Markdown, peak | 190.67 MiB | 225.06 MiB | +34.38 MiB (+18.0%) |

The scaled gap is much larger than a second source-text copy. Tree, injection facts, semantic caches, IPC buffers, and allocator retention all contribute. Latency fields from this collector are not performance evidence because macOS footprint inspection perturbs the target.

## Decision

Keep one worker, but leave the ADR memory gate open. Add document-scoped derived-state accounting and pressure-driven eviction/admission before legacy removal.

## Validation

- python3 -m unittest discover -s benches/profile -p test_*.py (117 passed)
- python3 -m py_compile benches/profile/collect_worker_memory.py benches/profile/drive.py
- six order-reversed release batches for each retained workload
- JSON validation for both retained result files

### DIFF
--- a/benches/profile/collect_worker_memory.py
+++ b/benches/profile/collect_worker_memory.py
@@ -1,0 +1,307 @@
+#!/usr/bin/env python3
+"""Measure aggregate parent-plus-worker RSS for paired LSP workloads."""
+
+import argparse
+import json
+import os
+import pathlib
+import platform
+import statistics
+import subprocess
+import sys
+import tempfile
+import threading
+import time
+import re
+
+from collect_worker_proxy import (
+    artifact_identity,
+    build_driver_command,
+    controlled_environment,
+    cpu_model,
+    harness_identity,
+    parse_driver_summary,
+    portable_environment,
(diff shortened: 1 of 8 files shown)


atusy:

PR #879, opened 2026-07-20
perf(worker): trim auxiliary caches under pressure

atusy:
## Summary

- retain exact-version semantic results while trimming recomputable injection-derived state once a document semantic cache exceeds 2 MiB
- extend the process-tree memory harness to compare two authoritative binaries across multiple simultaneously open documents
- record order-balanced memory and cache-hit latency evidence in the ADR

## Measurement

Across three simultaneous 1.07 MiB injection-heavy Markdown documents (six reversed-order batches):

| Metric | Stage 15 | Stage 16 | Change |
| --- | ---: | ---: | ---: |
| Stable physical footprint | 488.99 MiB | 472.17 MiB | -16.82 MiB (-3.4%) |
| Peak physical footprint | 513.13 MiB | 494.60 MiB | -18.52 MiB (-3.6%) |
| Stable aggregate RSS | 531.17 MiB | 519.06 MiB | -12.11 MiB (-2.3%) |

Unsampled exact-version latency remained within run-to-run variation (p50 -0.9%, p95 +0.8%).

This is intentionally a soft cache policy. Worker-wide derived-state accounting, deterministic global eviction, and a hard admission contract remain open ADR gates.

## Validation

- `cargo fmt --all -- --check`
- `cargo check --lib --tests --features e2e`
- `cargo test --lib` (3,132 passed, 2 ignored)
- `cargo test --features e2e --test tree_worker_process` (4 passed)
- `cargo test --features e2e --test e2e_tree_worker_shadow` (37 passed)
- profile Python tests (119 passed)

PR #880, opened 2026-07-20
feat(worker): bound retained parser memory

atusy:
## Summary

- account non-evictable document state, reusable result caches, and recomputable auxiliary caches across the resident worker
- deterministically evict derived caches above a 128 MiB soft budget
- transactionally reject source/host-tree growth above an independent 512 MiB estimated-memory bound
- skip worker-wide pressure scans on exact-version semantic cache hits
- record order-reversed release A/B measurements against Stage 16

## Measurement

Three simultaneous 1.07 MiB / 5,000-injection Markdown documents, six alternating-order batches:

| Metric | Stage 16 | Stage 17 | Change |
| --- | ---: | ---: | ---: |
| Stable physical footprint | 463.75 MiB | 472.49 MiB | +1.9% |
| Peak physical footprint | 493.99 MiB | 491.89 MiB | -0.4% |
| Stable aggregate RSS | 545.77 MiB | 533.81 MiB | -2.2% |
| Peak aggregate RSS | 546.73 MiB | 541.74 MiB | -0.9% |

The mixed deltas are run-to-run noise, not a memory-reduction claim. This stage establishes deterministic accounting and admission. Tree-sitter does not expose native allocation size, so the 512 MiB contract is defined over a conservative node-weight estimate rather than RSS.

Exact-version cache-hit latency, six order-reversed unsampled runs per binary:

| Metric | Stage 16 | Stage 17 | Change |
| --- | ---: | ---: | ---: |
| p50 | 75.45 ms | 76.25 ms | +1.1% |
| p95 | 79.85 ms | 80.30 ms | +0.6% |

These differences are below run-to-run variation.

## Validation

- `cargo test --lib`
- `cargo test --features e2e --test tree_worker_process`
- `cargo test --features e2e --test e2e_tree_worker_shadow`

## Stack

- base: #879
- ADR: #862

### DIFF
--- a/benches/profile/results/single_worker_stage17_memory_admission_cache_hit_2026-07-20.json
+++ b/benches/profile/results/single_worker_stage17_memory_admission_cache_hit_2026-07-20.json
@@ -1,0 +1,51 @@
+{
+  "experiment": "single worker stage 17 memory admission exact-version cache hit",
+  "date": "2026-07-20",
+  "baseline": {
+    "stage": 16,
+    "binary_sha256": "810f3a43b16002eebdb15a30f0da765131d485c82760894aa18b69790decca78"
+  },
+  "candidate": {
+    "stage": 17,
+    "binary_sha256": "fb44ffe210efa84f088914bc38e81e0b98120789f4c066d709d6cef77eb6c57d"
+  },
+  "scenario": {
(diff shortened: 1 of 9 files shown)


atusy:

PR #881, opened 2026-07-20
feat(worker): stress memory admission boundaries

atusy:
## Summary

- inject immutable, child-specific derived-cache and non-evictable memory budgets into the resident worker while preserving the 128 MiB / 512 MiB defaults
- expose stateless internal document-memory inspection and process tests for soft eviction, full-sync and incremental rollback, recovery, and worker isolation
- add a normal-release boundary stress driver and collectors, then record reduced-budget and unchanged-default A/B evidence in the ADR

## Why

Stage 17 established deterministic accounting and admission, but its production-sized defaults made the pressure boundary impractical to exercise end to end. This stage makes both limits testable at the real parent/worker process boundary without ambient environment overrides or allocating hundreds of MiB.

## Measurement

Six reduced-budget release runs all observed:

- eviction of the older derived result while preserving resident tree/node identity
- byte-identical recomputation
- transactional rejection of oversized full sync and incremental edit
- a successful smaller recovery edit in the same worker generation

With unchanged defaults, 12 order-reversed cache-hit runs showed no material latency change (p50 -0.5%, p95 -0.4%, wall/cycle -1.1%). Three-document memory medians were 12.44–16.80 MiB higher, but paired ranges crossed zero and were much wider than the median differences, so the result is recorded as noisy rather than a repeatable regression.

This validates the admission policy, not a total process-RSS cap.

## Validation

- `cargo test --lib` (3,146 passed, 2 ignored)
- `cargo test --features e2e --test tree_worker_process` (9 passed)
- `cargo test --features e2e --test e2e_tree_worker_shadow` (37 passed)
- profile Python tests (120 passed)

## Stack

Depends on #880.

### DIFF
--- a/benches/profile/collect_worker_memory_stress.py
+++ b/benches/profile/collect_worker_memory_stress.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python3
+"""Collect repeated reduced-limit worker memory-admission stress runs."""
+
+import argparse
+import json
+import os
+import pathlib
+import platform
+import subprocess
+import sys
+import threading
+import time
+
+from collect_worker_memory import (
+    descendant_pids,
+    read_process_table,
+    read_total_footprint_bytes,
+    summarize_samples,
+)
+from collect_worker_proxy import (
+    artifact_identity,
+    controlled_environment,
+    cpu_model,
+    portable_environment,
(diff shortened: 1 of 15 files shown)


atusy:

PR #882, opened 2026-07-20
feat(worker): converge systemic failures with a breaker

atusy:
## Summary

- stop spawning tree workers after three systemic worker-generation failures
- admit exactly one configuration-fenced half-open worker after cooldown
- keep parser-independent LSP service and shutdown available while the tree tier is degraded
- record repeated breaker E2Es and Stage 18 vs Stage 19 release A/B measurements

## Measurement

- breaker convergence: 3/3, median 21.564 s full E2E
- failed half-open: 3/3, median 23.249 s full E2E
- successful resync/close: 3/3, median 23.806 s full E2E
- cache-hit median paired deltas: p50 +0.3%, p95 +1.0%, wall +1.4%
- cold-start median paired deltas: elapsed +0.8%, first request +0.7%

## Validation

- `cargo fmt --all -- --check`
- `cargo test --lib` (3,147 passed, 2 ignored)
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --features e2e --test tree_worker_process` (9 passed)
- each new breaker E2E passed in all focused and measurement runs

## Stack

Depends on #881. This remains draft while later ADR stages and the full revise pipeline are in progress.

PR #883, opened 2026-07-20
fix(worker): terminate owned Unix process groups

atusy:
## Summary

- spawn the resident Unix worker as an explicitly owned process group
- terminate descendants on failed handshake, timeout, transport failure, drop, and graceful shutdown
- prove real LSP → worker → five-minute descendant cleanup without natural-expiry false positives
- record Stage 19 vs Stage 20 cold-start A/B and remaining platform lifecycle gaps

## Measurement

- descendant cleanup E2E: 3/3, median 4.079 s full scenario
- cold-start median paired deltas: driver -0.4%, first request -0.6%

## Validation

- `cargo test --lib` (3,148 passed, 2 ignored)
- `cargo test --features e2e --test tree_worker_process` (9 passed)
- descendant cleanup focused and repeated E2Es

## Boundary

This closes explicit Unix cleanup only. Parent-death liveness and Windows Job Object ownership remain open ADR gates.

## Stack

Depends on #882. This remains draft while later ADR stages and the full revise pipeline are in progress.

### DIFF
--- a/benches/profile/results/single_worker_stage20_descendant_cleanup_2026-07-20.json
+++ b/benches/profile/results/single_worker_stage20_descendant_cleanup_2026-07-20.json
@@ -1,0 +1,51 @@
+{
+  "schema": 1,
+  "experiment": "single-tree-worker-stage20-descendant-cleanup",
+  "date": "2026-07-20",
+  "environment": {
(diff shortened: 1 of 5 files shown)


atusy:

PR #884, opened 2026-07-20
fix(worker): exit when the Unix parent dies

atusy:
## Summary

- add a close-on-exec Unix parent-liveness pipe to every resident worker
- monitor EOF from a dedicated control thread independent of the compute pool
- use `_exit` so parent death cannot join a permanently hung compute thread
- record repeated parent-death E2Es and Stage 20 vs Stage 21 cold-start A/B

## Measurement

- hung-compute parent-death E2E: 3/3, median 3.923 s full scenario
- cold-start median paired deltas: driver +0.15%, first request +0.05%

## Validation

- rustfmt
- process-group lifecycle unit tests
- repeated parent-death E2E

## Boundary

This closes the portable Unix/macOS liveness-pipe path. Linux PDEATHSIG hardening and Windows parent-handle/Job Object ownership remain open.

## Stack

Depends on #883. This remains draft while later ADR stages and the full revise pipeline are in progress.

### DIFF
--- a/benches/profile/results/single_worker_stage21_parent_liveness_2026-07-20.json
+++ b/benches/profile/results/single_worker_stage21_parent_liveness_2026-07-20.json
@@ -1,0 +1,24 @@
+{
+  "schema": 1,
+  "experiment": "single-tree-worker-stage21-parent-liveness",
+  "date": "2026-07-20",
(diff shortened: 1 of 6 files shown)


atusy:

PR #885, opened 2026-07-20
fix(worker): harden Linux parent-death handling

atusy:
## Status

GREEN and measured. Linux kernel parent-death delivery and the registration race are both proven independently of the portable liveness pipe.

## Scope

- pass one validated immutable liveness FD / expected-parent PID startup contract
- arm `PR_SET_PDEATHSIG(SIGKILL)` before observable worker work
- verify the armed signal with `PR_GET_PDEATHSIG` in the Linux E2E topology
- close parent-loss-before-registration with an immediate PID comparison
- retain the Stage 21 portable Unix liveness pipe as fallback
- compile all targets/features on Ubuntu, macOS, and Windows

## Measurement

Twelve order-reversed Ubuntu release pairs compared immutable Stage 21 and Stage 22 binaries with matching-generation harnesses.

| Metric | Stage 21 median | Stage 22 median | Median paired delta |
| --- | ---: | ---: | ---: |
| Worker handshake | 80.002 ms | 80.024 ms | -0.028% |
| LSP process through clean shutdown | 273.915 ms | 274.118 ms | +0.430% |
| First semantic request | 101.556 ms | 101.737 ms | +0.284% |

No startup regression or improvement was detected. Raw values and binary digests are committed under `benches/profile/results/`.

## Validation

- Linux PDEATHSIG delivery test: pass
- deterministic registration-race test: pass
- `cargo check --all-targets --all-features`: Ubuntu/macOS/Windows pass
- Check, Test, Clippy, Format, Audit, benchmark tooling: pass

## Stack

Depends on #884. Remains draft until stack-wide revise; Windows ownership and later native hazard/quarantine, protocol, compatibility, and legacy-removal stages remain open.


atusy:

PR #886, opened 2026-07-20
fix(worker): own Windows worker process trees

atusy:
## Summary

- gate Windows worker startup until the parent transfers a validated process handle and assigns a kill-on-close Job Object
- supervise parent death from a dedicated thread that hard-terminates even when compute completion is hung
- keep the Job handle parent-only and fail closed on every partial-spawn error
- prove parent-monitor-only, Job-only, and combined behavior with stable process handles
- retain a dedicated Windows lifecycle job and version the paired startup measurement

## Correctness evidence

- Initial run `29748705954` was invalid because unrelated parser fixture setup prevented the intended hang.
- Corrected RED run `29750012782` reached the lifecycle assertion and failed because the worker descendant survived parent death.
- GREEN run `29751887031` passed parent-monitor-only, Job-only, and combined Windows cases, plus Windows/macOS/Ubuntu cross-platform checks.

## Performance

Windows Server 2025, 5 warmup + 30 order-reversed pairs, four worker threads, parser-free:

| Metric | Stage 22 | Stage 23 | Paired median |
| --- | ---: | ---: | ---: |
| HandshakeReady | 73.649 ms | 74.023 ms | +0.537% |
| Clean shutdown | 5.283 ms | 5.298 ms | +0.171% |

The Windows protection adds a small, repeatable median startup cost of 0.396 ms and no detected shutdown change. Raw values and harness digests are versioned in the PR.

Stacked on #885.


atusy:

PR #887, opened 2026-07-20
feat(worker): acknowledge active grammar hazards

atusy:
## Summary

- add a protocol-17 acknowledged grammar-hazard lease before grammar-backed work
- retain acknowledged leases in the parent for exact in-session crash attribution
- prove post-ack crash quarantine and healthy-grammar recovery end to end
- record 24 order-reversed Stage 23/24 release pairs and the resulting ADR decision

## Correctness

- cargo check --all-targets --all-features
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test --lib tree_worker::tests -- --quiet (61 passed)
- focused crash E2E: acknowledged Rust hazard logged and quarantined, worker restarted and resynced, healthy Lua grammar recovered

## Measurement

On macOS 26.5 / Apple M4, 24 paired runs with 500 Rust requests and 4 worker threads measured:

- sequential p50: +5.382%, +77 microseconds paired median; Stage 24 slower in 20/24 pairs
- arm/ack queue wait: increased in 24/24 pairs
- four-way throughput: +3.898%; Stage 24 improved in 20/24 pairs
- four-way p95: -10.021%; Stage 24 improved in 20/24 pairs

The ACK acts as accidental admission pacing under concurrency, but adds a repeatable synchronous hot-path cost. This stage is the correctness reference and does not pass the fine-grained latency gate.

## Follow-up boundary

This stage leases the request host grammar and quarantines only the first surviving lease. Runtime injection identities, complete multi-hazard quarantine, independently bounded control transport, and native-segment deadlines remain later stages. The next stage must remove the synchronous pipe round trip while separately preserving or replacing the measured admission-pacing effect.

Stacked on #886.

### DIFF
--- a/benches/profile/results/single_worker_stage24_grammar_hazard_2026-07-20.json
+++ b/benches/profile/results/single_worker_stage24_grammar_hazard_2026-07-20.json
@@ -1,0 +1,2819 @@
+{
+  "baseline": {
+    "binary_sha256": "b5ef5e8b4b39a84433fd6c337f1fd90af7dcf2324f5d76d6270083ce77c3c9c1",
+    "commit": "82f4b9218817b82930ad9f15fed37694bfb35988",
+    "harness_sha256": "e76fac356bd426a0b175f649fcb6c9497673175bdd28f897ddd19adef5a85be4",
+    "stage": 23
+  },
+  "candidate": {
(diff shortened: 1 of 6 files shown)


atusy:

PR #888, opened 2026-07-20
perf(worker): commit hazard frames before native entry

atusy:
## Summary

- replace the grammar-hazard parent acknowledgment with a worker-local writer receipt
- enter grammar-backed work only after the complete arm frame is committed to the OS pipe
- preserve arm-before-release/EOF crash attribution while removing protocol ACK and waiter state
- measure the sequential latency recovery separately from the parallel pacing removed with the ACK

## TDD and correctness

RED: an E2E worker aborted on any Stage 24 hazard ACK; the semantic response took 16.72 seconds and failed a ten-second latency bound.

GREEN:

- no hazard ACK exists in protocol 18
- the same request completes within the bound without worker restart
- post-commit crash E2E still logs and quarantines the exact Rust artifact, restarts/resyncs, and serves healthy Lua
- cargo check --all-targets --all-features
- cargo clippy --all-targets --all-features -- -D warnings
- 61 focused worker unit tests pass

## Measurement

On macOS 26.5 / Apple M4, 24 paired Stage 24/25 runs with 500 Rust requests and 4 worker threads measured:

- sequential p50: -3.300%; Stage 25 improved in 21/24 pairs
- arm wait p50: -46.909%; Stage 25 improved in 24/24 pairs
- four-way throughput: -3.151%; Stage 25 regressed in 20/24 pairs
- four-way p95: +9.200%; Stage 25 regressed in 20/24 pairs

The removed parent round trip was both hot-path overhead and accidental admission pacing. This PR keeps crash ordering independent of scheduling; a later stage must test explicit workload-aware admission without reintroducing ACK or fixed sleeps.

Stacked on #887.

### DIFF
--- a/benches/profile/results/single_worker_stage25_committed_hazard_2026-07-20.json
+++ b/benches/profile/results/single_worker_stage25_committed_hazard_2026-07-20.json
@@ -1,0 +1,2819 @@
+{
+  "baseline": {
+    "binary_sha256": "f64e66521e12e5e19d51454833cb8526bc6a404fb11594a290e3d2906d3627fd",
+    "commit": "f8b647522",
+    "harness_sha256": "052bca0015337cb7bfdef7bd4cef14f67eb2bef09c218c0c55b5a8a234f7ab09",
+    "stage": 24
+  },
+  "candidate": {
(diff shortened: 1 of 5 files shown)


atusy:

PR #890, opened 2026-07-21
test(semantic): measure unique typing states

atusy:
## Summary

- measure semantic-token typing through unique document states and queued/cancelled bursts
- reconstruct full and delta responses client-side instead of trusting `resultId`
- reject a timing sample unless the tracked token position matches the latest edit
- optionally retain raw per-request samples and report p95

## Why

A rotating `resultId` does not prove that token content follows the current document. Likewise, toggling between two document states can hit the whole-document cache and understate real typing work. This stage establishes a latest-edit correctness and latency contract before scheduling changes.

`latest-wins` here is admission, not trailing-edge debounce: the newest version remains immediately eligible, while only work whose result is already unusable may be cancelled.

## Stack

- base: #889
- this PR: Stage 27 typing measurement contract

## Validation

- Rust and Markdown single-edit and eight-edit burst scenarios reconstruct and verify current tokens
- cancellation burst verifies the final full result against every preceding edit
- CI

### DIFF
--- a/benches/semantic_tokens.rs
+++ b/benches/semantic_tokens.rs
@@ -423,8 +423,14 @@ fn summarize(samples: &[Duration]) -> Stats {
     let mut samples = samples.to_vec();
     samples.sort_unstable();
     let pick = |q: f64| samples[((samples.len() as f64 * q) as usize).min(samples.len() - 1)];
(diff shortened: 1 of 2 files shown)


atusy:

PR #891, opened 2026-07-21
chore(worker): expose internal debug telemetry

atusy:
## Summary

- initialize the hidden tree worker's logger from `RUST_LOG`
- keep worker diagnostics on inherited stderr; stdout remains protocol-only
- expose existing semantic phase timing from inside the isolation boundary

## Why

Parent metrics report queue and round-trip time but cannot separate host query, injection, and finalization costs owned by the worker. The existing phase logs were unreachable because the hidden worker never initialized a logger. This stage makes performance diagnosis possible without enabling logs or adding hot-path work by default.

## Stack

- base: #890
- this PR: Stage 28 worker telemetry

## Validation

- `cargo clippy --bin kakehashi --all-features -- -D warnings`
- authoritative semantic typing run with `RUST_LOG=kakehashi::semantic=debug,kakehashi::tree_worker_shadow_metrics=debug`

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -363,26 +363,35 @@ fn main() -> ExitCode {
             non_evictable_estimate_hard_bytes,
             parent_liveness_fd,
             expected_parent_pid,
-        }) => kakehashi::tree_worker::WorkerMemoryBudgets::new(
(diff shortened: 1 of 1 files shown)


atusy:

PR #893, opened 2026-07-22
bench(semantic): attest rejected query metadata experiment

atusy:
## Summary\n\n- record Stage 30 semantic-query metadata memoization as rejected after provenance-hardened reruns\n- restore the production semantic-token collector exactly to Stage 29\n- add sparse 32/64 KiB benchmark controls\n- hash each executed server binary before measurement and store the SHA-256 in every raw binary/run record\n- retain four alternating-order raw pairs, an aggregate, and the rejection ADR\n\n## Decision\n\nDo not ship request-local query metadata memoization.\n\nThe cryptographically attested rerun did not demonstrate a reliable primary typing-path improvement:\n\n- Rust single-edit typing median: +10.0% (range -26.0% to +28.7%)\n- Rust eight-edit burst median: +2.0% (range -15.6% to +9.1%)\n- Rust cancellation burst median: -12.0% (improved in all four pairs)\n- sparse Rust 64 KiB median: +11.6% (regressed in all four pairs)\n\nThe cancellation win does not justify the consistent sparse regression and inconclusive primary typing results.\n\n## Measurement integrity\n\n- content-addressed Stage 29 and Stage 30 binaries\n- runtime SHA-256 recorded in each raw manifest and run\n- four alternating binary-order pairs\n- 6 warmups and 30 timed iterations per binary/scenario\n- 2,160 latest-token responses validated\n- 1,152 obsolete requests validated as RequestCancelled\n\n## Validation\n\n- cargo test --lib: 3180 passed, 2 ignored\n- cargo test --lib analysis::semantic::token_collector::tests\n- cargo bench --bench semantic_tokens --features e2e --no-run\n- cargo clippy --all-targets --all-features -- -D warnings\n- raw binary attestations, raw hashes, and aggregate statistics independently recomputed\n- same-thread and fresh Codex reviews: CLEAN\n- independent performance review: CLEAN\n\nStacked on #892.

PR #894, opened 2026-07-22
docs(worker): record rejected Stage 31 finalize fast paths

atusy:
## Outcome

**Rejected after review and remeasurement; do not merge.**

Stage 31 tried to preserve the input token vector when semantic finalization had no `@none` captures or multiline tokens. Local review found that the new `@none` probe could extend cancellation latency, so the final candidate polls once per 64-token chunk and has a discriminating mid-scan cancellation test.

## Final authoritative measurement

Four alternating authoritative-worker pairs, each with 6 warmups and 30 timed iterations per binary/scenario, measured the cancellation-safe candidate against Stage 30:

- Rust single-edit typing median: **+0.8%** (2/4 improved; no improvement).
- Rust eight-edit burst median: **+1.6%** (2/4 improved; no improvement).
- Rust cancellation burst median: **-9.0%** (all 4 improved).
- Markdown single-edit central median: **-1.0%**, but p95 regressed in all 4 pairs (**+32.9%** median).

The consistent cancellation-follow gain does not justify the lack of normal Rust typing improvement and the Markdown tail regression. The branch retains the final 1,200-sample series, exact source/binary hashes, and four raw pairs for an earlier rejected disjoint-line shortcut.

Evidence: `docs/architecture-decisions/single-resident-multithreaded-tree-worker-stage31-measurement.md`

## Validation

- 215 semantic tests
- focused allocation identity and mid-scan cancellation tests
- JSON and SHA-256 validation of all 8 raw runs

Stacked on #893 only to preserve the experiment and its evidence.


atusy:
Local revise and authoritative remeasurement converged on rejection. The cancellation-safe candidate improved cancel-burst central latency in all four pairs (-9.0%), but did not improve normal Rust typing (+0.8% single, +1.6% burst) and regressed Markdown single-edit p95 in all four pairs (+32.9% paired median). Closing rather than merging a benchmark-specific tradeoff. The branch retains the final and rejected-candidate raw evidence.

Issue #895, opened 2026-07-22
Improve semantic-token performance without a tree worker


atusy:
## Context

The single-tree-worker effort was introduced to isolate the LSP process from native parser and external-scanner failures. We have not observed such a failure as a practical production problem, so the isolation architecture is not justified by current evidence. The worker task is intentionally retired rather than deferred; the worker PR stack #862 through #893 was closed by decision.

The experiments nevertheless exposed semantic-token improvements that do not inherently require a child process. A fresh non-worker comparison supports extracting those improvements independently.

### Exploratory non-worker evidence

In a local exploratory run on Apple M4, release binaries from `origin/main` (`a1278be5`) and the final experiment stack (`f3febbf4`) were compared with `KAKEHASHI_TREE_WORKER_MODE=legacy`, so no child worker handled tree operations. Four alternating-order pairs used 6 warmups and 30 timed iterations per binary/scenario. Every typing response was validated against the latest unique edit state; obsolete cancellation-burst requests had to return `RequestCancelled`.

Median paired deltas for the non-worker experiment stack relative to `origin/main` were:

- Rust large single-edit typing: **-1.8%** (range -5.5% to +1.0%; no broad win claim)
- sparse Rust 32 KiB controls: **-45.5% to -47.6%**, improved in every pair
- sparse Rust 64 KiB: **-47.5%**, improved in every pair
- Markdown injection typing: **-21.2%**, improved in every pair
- Markdown eight-edit burst: **-10.1%**, improved in every pair
- Rust eight-edit burst: **+3.6%**, regressed in every pair
- Rust cancellation burst: **-1.3%** (range -13.6% to +15.4%; inconclusive)

These numbers indicate that the accumulated stack can improve some non-worker scenarios, but attribution and extractability remain unproven: the legacy binary still contains the accumulated experiment-stack source. The raw files currently exist only as local temporary artifacts and are not durable evidence. The first benchmark-contract PR must reproduce and archive the comparison with source, binary, harness, fixture, environment, and sample attestations. Each extraction below must stand on its own measurement.

## Goal

Improve semantic-token latency and contention behavior in the existing in-process architecture, without introducing a worker process, IPC protocol, duplicated document/tree state, or a worker-based parser-crash isolation task. Existing in-process crash-marker correctness work (#725, #726, and #727) is independent and out of scope for this issue.

[...]

PR #901, opened 2026-07-22
perf(semantic): avoid legend lookup allocation

atusy:
## Summary

- replace the per-capture `split().collect::<Vec<_>>()` legend lookup with direct iterator consumption
- pin empty, repeated, and trailing modifier behavior before changing the hot path
- retain exact before/after benchmark evidence plus an independent higher-confidence confirmation run

This is stacked on #900 and advances the low-state hot-path work tracked by #895. It does not change scheduling, cancellation, caching, or semantic freshness.

## Performance hypothesis

Every semantic capture lookup previously materialized a temporary `Vec<&str>`. The new implementation consumes `split` directly, removing that allocation while preserving token type and modifier semantics.

Exact measured sources:

- A: `benchmark/semantic-legend-before-2026-07-23` (`22c5467f7cf2270c35f4088fdb8bd0d7dd55af36`)
- B: `benchmark/semantic-legend-after-2026-07-23` (`ae8a7c1866888cb88c39850212c5e06383a6c8de`)
- harness: B (`ae8a7c1866888cb88c39850212c5e06383a6c8de`)

Both durable tags are published on `origin`.

## Benchmark results

Default comparison: 4 alternating AB/BA pairs, 6 warmups, 30 retained samples, all 12 scenarios. Evidence: `benches/results/semantic-legend-allocation-2026-07-23/`.

- Markdown typing burst: +6.55% median, paired range +0.53%..+11.75%
- Rust typing burst: +3.15%, range -5.09%..+15.95%
- Rust sparse 64 KiB: -4.57%, range -9.55%..+3.59%
- Remaining scenarios were mixed and mostly crossed zero

Because the apparent Markdown regression was suspicious, I ran an independent confirmation with 8 alternating AB/BA pairs, 12 warmups, 30 samples, and 9 representative scenarios. Evidence: `benches/results/semantic-legend-allocation-confirmation-2026-07-23/`.

- Markdown typing burst: -4.63%, range -14.98%..+2.95%
- Rust typing burst: -6.69%, range -11.71%..+9.73%
- Rust typing delta: -5.75%, range -15.96%..+13.50%
- Rust cancel burst: -1.63%, range -8.64%..-0.98%
- Other representative scenarios crossed zero

The initial Markdown regression did not reproduce and reversed direction. The honest conclusion is that this change deterministically removes one temporary `Vec` allocation per capture lookup, but its whole-LSP latency effect is below the current run-to-run variance. This PR therefore makes no broad end-to-end speedup claim and shows no reproducible regression.

## Semantic compatibility

Characterization covers:

- empty base with modifier remains unmapped
- empty interior and trailing modifier components remain ignored
- duplicate modifiers remain idempotent
- unknown modifiers remain ignored

## Validation

- `cargo test --quiet`: 3043 passed, 2 ignored; integration and doc tests passed
- `make check`
- `cargo test map_capture_to_token_type_and_modifiers`: 29 passed
- `cargo clippy -- -D warnings`
- `KAKEHASHI_BENCH_VALIDATE_SCENARIOS=1 cargo test --bench semantic_tokens --features e2e`
- `python3 -m unittest discover -s benches -p test_*.py`: 18 passed
- `cargo test --test semantic_benchmark_contract --features e2e`: 9 passed
- raw/stdout/summary/fixture-manifest hashes rechecked against both manifests
- local multi-perspective review and two-stage fresh Codex MCP review converged with no remaining findings

## Stack

1. #899 benchmark contract
2. #900 synchronous stateless semantic core
3. this PR: allocation-free builtin legend lookup

Please review and merge in stack order.

---

**Note (2026-08-11): raw samples removed**

`benches/results/*/pair-*.json` (the per-pair raw sample arrays) were removed from every PR in this stack so that raw benchmark data does not enter the repository. The evidence directories remain and still carry `summary.json` (per-pair median / delta / min / max), `pair-*.txt` (the per-scenario A/B table including p95), and `manifest.json` / `fixture-manifest.json` (commits, tags, SHA-256).

**Every number quoted above is still reconstructible from these.** Only the 30-sample raw arrays are gone. `manifest.json` deliberately keeps `raw_files` / `samples_sha256` as a record of what was dropped, and `.gitignore` now carries `benches/results/*/pair-*.json` to prevent re-introduction.


atusy:
@codex review

atusy:
@codex review

atusy:
## Why this is being closed

This PR's own conclusion already states the effect is below run-to-run variance, and the measurement against current `main` does not contradict it. The 8-line change is harmless but has no demonstrated value, and the #900 seam it sits on is also being closed.

## Measured against current `main` (2026-08-11)

Collected with `benches/collect_semantic_pairs.py`: A = `origin/main` `9c4de0d2f`, **8 alternating AB/BA pairs, 12 warmups, 30 retained samples**, 14 scenarios. A difference counts only when every pair agrees in sign.

| B | decidable difference |
|---|---|
| #899 + #900 | **none** (all 14 scenarios cross zero) |
| ↑ + #913 | WIN `rust_large/stable_edit_followup_delta` 51.35 → 31.91 ms (**−19.44 ms**, −42.8%) only |
| #903 | WIN `rust_xlarge/cancel_burst` 136.47 → 123.41 ms (**−13.06 ms**, −9.9%) / REG `rust_large/typing_burst` +0.39 ms |

**Every sparse scenario crossed zero in all three runs.** None of these PRs captures the 39–47% sparse headroom recorded in #895.

Note: several regressions that looked unanimous at 4 pairs disappeared at 8 (e.g. #913's `full_cache_hit` +15.3% → −1.9%). Four pairs is too weak a criterion — the same effect is documented in #901's own body.

The branch is not deleted, so this can be restored if the work resumes.

Issue #896, opened 2026-07-22
Explore revision-scoped incremental semantic-token artifacts


atusy:
## Context

#895 tracks measurable semantic-token improvements in the existing in-process
architecture. It also identifies changed-range reuse and snapshot-owned lazy
derived data as high-upside experiments, but intentionally keeps its delivery
policy focused on small, independently positive PRs.

The current architecture already has the right correctness foundation for a
larger change: document inputs are versioned, and parsing publishes immutable,
internally consistent `ParseSnapshot`s. On a semantic cache miss, however, the
server still reconstructs a flat whole-document token result. A small edit can
therefore pay whole-document host query, injection merge/finalization, token
encoding, and delta comparison costs even when most syntax is structurally
unchanged.

This issue explores extending the immutable-revision model through the semantic
pipeline rather than adding another mutable ownership model.

## Goal

Make semantic work proportional to the semantic area affected by an edit when
correctness can be proved, while retaining exact full-computation fallback.

The intended shape is:

```text
Document revision
  -> ParseSnapshot
  -> Injection/layer forest
  -> Layer-local raw-token chunks
  -> Overlap-resolved token tree
  -> full / delta / range views
```

Each artifact must be immutable and identified by document incarnation, parsed
content version, semantic settings/query generation, and any session-constant
client capability that affects token production. Derived work may be lazy,
single-flighted, or prefetched, but a partial or obsolete artifact must never be
published as current.

## Decision drivers

- Preserve the latest-edit semantic freshness contract from #895.
- Prefer immutable revision state over a text-owning document actor or mutable
  cross-revision token table.
- Reduce asymptotic work for large documents with small edits.
- Share identical-snapshot work across full, delta, range, refresh-driven, and
  concurrent requests without coalescing edits.
- Bound retained state by explicit document and byte budgets.
- Keep a byte-for-byte full-computation oracle until every incremental safety
  class is proven.

### 1. Revision-scoped `SemanticArtifact`

Introduce one immutable artifact interface associated with `ParseSnapshot`.
Initially it may wrap the existing whole-document result unchanged. Requests for
the same snapshot/generation join one computation and independently shape full,
delta, or range responses.

The request remains responsible for final currency validation. A newer document
revision cancels obsolete computation; multiple consumers of the current revision
do not cancel each other.

### 2. Persistent chunked token representation

Represent pre-finalize and finalized tokens as ordered chunks, such as a
[...]

PR #911, opened 2026-07-23
test(bench): measure same-snapshot semantic fan-out

atusy:
## Summary

- add a whole-LSP benchmark for concurrent delta and range consumers of one freshly parsed snapshot
- positively observe the edited parse through the ingress-ordered `kakehashi/node` endpoint before timing
- use opt-in benchmark instrumentation to rendezvous after both completed-artifact cache lookups miss
- validate the delta baseline and the complete independently projected range payload

## Why

Issue #896 proposes sharing one immutable semantic artifact across request-local full, delta, and range responses. Existing cache-hit and typing scenarios do not prove that two concurrent consumers reached the same uncached snapshot. The benchmark-only barrier makes the contract deterministic: the current implementation releases two producers, while Stage 3 will release one producer and one in-flight joiner from the same point.

## Benchmark contract

- scenario: `rust_xlarge/same_snapshot_fanout`
- parse latency is outside the timed section
- timed value is all-consumer latency for delta + range
- collect separately with 4 alternating AB/BA pairs, 6 warmups, and 30 retained samples
- collector enables and attests `semantic-bench-instrumentation` only for this isolated scenario
- production default scenarios remain feature-free; mixed or empty filters are rejected
- direct runs fail explicitly if the synchronization feature is absent

## Validation

- `cargo fmt --check`
- `git diff --check`
- `python3 -m unittest test_collect_semantic_pairs.py` (15 passed)
- `cargo clippy --lib --bin kakehashi -- -D warnings`
- `cargo clippy --lib --bin kakehashi --features semantic-bench-instrumentation -- -D warnings`
- `KAKEHASHI_BENCH_VALIDATE_SCENARIOS=1 cargo bench --bench semantic_tokens --features e2e --no-run`
- release smoke with instrumentation: 1 warmup + 3 retained samples; all complete delta/range freshness checks passed

A focused unit-test link was attempted, but the machine ran out of disk while linking test artifacts. This was an environment failure after the clippy, benchmark build, Python tests, and release smoke checks above had passed.

## Dependencies

- based on #899 so the measurement contract can precede the later #896 implementation PR
- no parser libraries or query files are committed; fixtures are installed through kakehashi and only their manifest is retained

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - run: cargo check
+      - run: cargo check --features semantic-bench-instrumentation
       - run: cargo check --bench semantic_tokens --features e2e
       - run: KAKEHASHI_BENCH_VALIDATE_SCENARIOS=1 cargo test --bench semantic_tokens --features e2e
 
@@ -44,6 +45,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - run: cargo clippy -- -D warnings
+      - run: cargo clippy --features semantic-bench-instrumentation -- -D warnings
       - run: cargo clippy --bench semantic_tokens --features e2e -- -D warnings
 
   audit:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,7 @@ nix = { version = "0.31", features = ["signal", "process", "fs"] }
 
 [features]
 e2e = []
+semantic-bench-instrumentation = []
 
 # Unix-only dev dependencies: pipe() for deterministic broken-pipe tests,
 # Signal::SIGPIPE for asserting the termination signal.
--- a/benches/README.md
+++ b/benches/README.md
@@ -32,6 +32,13 @@ attestation, and raw-contract check succeeds. A failed run removes its staging
 artifacts and reports any exact worktree cleanup failure, so the same output
 path can be retried.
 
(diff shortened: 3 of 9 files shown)


atusy:
@codex review

atusy:
## Why this is being closed

The fan-out benchmark exists to make #912's contract measurable. With #912 closed there is nothing for it to measure, and it would otherwise leave 41 lines of benchmark-only instrumentation on a production path.

## Measured against current `main` (2026-08-11)

Collected with `benches/collect_semantic_pairs.py`: A = `origin/main` `9c4de0d2f`, **8 alternating AB/BA pairs, 12 warmups, 30 retained samples**, 14 scenarios. A difference counts only when every pair agrees in sign.

| B | decidable difference |
|---|---|
| #899 + #900 | **none** (all 14 scenarios cross zero) |
| ↑ + #913 | WIN `rust_large/stable_edit_followup_delta` 51.35 → 31.91 ms (**−19.44 ms**, −42.8%) only |
| #903 | WIN `rust_xlarge/cancel_burst` 136.47 → 123.41 ms (**−13.06 ms**, −9.9%) / REG `rust_large/typing_burst` +0.39 ms |

**Every sparse scenario crossed zero in all three runs.** None of these PRs captures the 39–47% sparse headroom recorded in #895.

Note: several regressions that looked unanimous at 4 pairs disappeared at 8 (e.g. #913's `full_cache_hit` +15.3% → −1.9%). Four pairs is too weak a criterion — the same effect is documented in #901's own body.

The branch is not deleted, so this can be restored if the work resumes.

Issue #348, opened 2026-06-10
Server crash (heap corruption) on didChange after parsing the markdown_inline injection layer


atusy:
## Summary

The server process dies (SIGTRAP via malloc freelist corruption) when a document is edited **after** the `markdown_inline` injection layer has been parsed through the injection-stack machinery. This pre-dates the captures work: it reproduces with only `kakehashi/node` + `didChange`.

## Minimal repro (e2e-style, node path only)

1. `didOpen` a markdown document, e.g. `# Title\n\n...`
2. `kakehashi/node` with `{ position: inside "Title", injection: true }` — resolves the deepest layer (`markdown_inline`), loading + parsing the inline layer via `injection_stack_at`
3. `textDocument/didChange` (full-text replacement is enough)
4. Any subsequent request → EOF; the server is gone

Step 2 with a position inside a **python** code block instead (so only the python layer is parsed) does **not** crash — the existing "injected node identity survives an edit" e2e passes. The lethal ingredient is the markdown_inline layer specifically.

## Crash evidence (macOS .ips)

- Faulting thread: panic-in-panic abort — `Stdout::write_all` hit `Mutex::lock::fail`, and the panic hook then trapped inside `_xzm_xzone_malloc_freelist_outlined` (heap corruption detected).
- Concurrent thread: `ParseCoordinator::parse_document` → `Parser::parse` (incremental host reparse) → `markdown.dylib tree_sitter_markdown_external_scanner_serialize` → `_platform_memmove` — i.e. the markdown **block** scanner serializing state during the didChange reparse while the heap is already corrupt.

## What it is NOT

- **Not a stale parser artifact**: deleting `deps/test/kakehashi` and letting `ensure_test_languages_installed` rebuild fresh dylibs reproduces identically.
- **Not an upstream grammar bug reachable via the bundled crates**: an in-process repro using `tree_sitter_md::LANGUAGE` + `tree_sitter_md::INLINE_LANGUAGE` (parse host → parse inline with `set_included_ranges` → `tree.edit()` → incremental reparse) passes cleanly. The crash needs the **installed dylibs** loaded via libloading.

## Suspicions

[...]

PR #921, opened 2026-07-29
refactor(parser): remove the parser crash quarantine

### LINKED ISSUE #725: crash recovery: persist parser activity before an uncatchable parser crash
## Summary

`FailedParserRegistry` claims to recover from parser crashes such as C assertion failures / `SIGABRT`, but the current implementation records active parsers only in memory. The `parsing_in_progress` file consumed on the next startup is written only from graceful shutdown or the SIGTERM/SIGHUP cleanup path. An uncatchable parser crash therefore leaves no durable marker, so the same parser is loaded again after restart.

## Evidence

In `src/language/failed_parsers.rs`:

- the module contract explicitly says the mechanism handles C assertion failures (`SIGABRT`) by recording the parser before parsing;
- `begin_parsing()` only increments `parsing_counts` in memory;
- `persist_state()` is the only method that writes `parsing_in_progress`.

The production call sites for `persist_state()` are lifecycle cleanup paths:

- `shutdown_impl()` (normal LSP shutdown);
- the SIGTERM/SIGHUP watcher.

Neither runs after `SIGABRT`, `SIGSEGV`, `SIGKILL`, `abort()`, or an equivalent hard process failure. On restart, `init()` therefore sees no `parsing_in_progress` file and cannot quarantine the crashing parser.

The existing crash tests do not exercise this contract: they explicitly call `persist_state()` between `begin_parsing()` and constructing the restarted registry, which simulates a graceful lifecycle callback rather than a crash.

This regression was introduced by `7d36519dd` when synchronous marker writes were removed from `begin_parsing()` to reduce didChange latency. The commit preserved startup reading but moved the only persistence to graceful shutdown, which cannot establish what was running immediately before an uncatchable crash.

## Impact

A malformed or buggy native tree-sitter parser can put the editor into a restart/crash loop. The documented safety property—quarantine the parser that killed the prior kakehashi process—is not provided for the failures it is intended to handle.

## Reproduction / regression test shape

A subprocess test can make the contract deterministic:

1. child process creates a registry and calls `begin_parsing("crasher")`;
2. child terminates without calling `persist_state()` or `end_parsing_language()` (ideally via `abort()` to model the documented failure);
3. parent constructs a new registry on the same state directory and calls `init()`;
4. expected: `crasher` is failed; current result: it is not failed.

A smaller unit test can demonstrate the same persistence gap by dropping the first registry immediately after `begin_parsing()` without calling `persist_state()`, though the subprocess version better pins the uncatchable-crash contract.

## Suggested direction

Restore a durable “currently parsing” marker before native parser execution and clear/update it after completion, without reintroducing blocking filesystem I/O on the async request path. For example, serialize marker updates through a dedicated blocking worker and require acknowledgement before entering native parser code. The ordering guarantee is essential: fire-and-forget persistence can still lose the marker if the parser crashes before the write executes.

Concurrent parser activity must remain represented (the current per-language counts allow more than one active language), and marker replacement/clearing should be crash-safe so a partial write cannot erase recovery evidence.

### LINKED ISSUE #726: crash recovery: isolate parser markers between concurrent server processes
## Summary

Parser crash-recovery state is shared as `<data-dir>/parsing_in_progress` across every kakehashi process using the default data directory. A newly starting server treats any existing marker as evidence that the prior process crashed, even when the marker belongs to another still-running editor session.

## Impact

Concurrent editor/server processes can poison each other’s parser registry. If one process has an active parse marker while another starts, the second can mark that healthy parser permanently failed and remove the first process's marker. With durable pre-parse markers (required to make hard-crash recovery actually work; #725), this race becomes a normal rather than shutdown-only window.

`KAKEHASHI_STATE_DIR` can isolate deliberately configured processes, but normal users share the default data directory and no per-process/session namespace is created automatically.

## Evidence

- `AutoInstallManager::init_failed_parser_registry()` defaults the state directory to the shared kakehashi data directory.
- `FailedParserRegistry::parsing_state_path()` returns the single fixed filename `parsing_in_progress`.
- `init()` consumes and deletes that marker without ownership/liveness information.

## Suggested direction

Use a per-process/session marker whose file remains exclusively locked for the lifetime of its owning registry. Startup can scan markers: a lock that cannot be acquired belongs to a live peer and must be ignored; a lock that can be acquired was released by a crashed process and its contents can be quarantined and removed. This avoids platform-specific PID-liveness checks and PID-reuse races. Marker content updates should stay ordered with parser entry/exit and crash-safe.

This should be covered by a test with two registries sharing one state directory: initializing the second while the first has an active marker must not mark the active language failed or consume the first registry's recovery evidence.

### LINKED ISSUE #730: tests: isolate crash recovery state from shared parser assets
## Summary

Unit-test builds currently derive `FailedParserRegistry` state from the shared `deps/` parser/query asset directory. The test helper clears only the legacy `parsing_in_progress` and `failed_parsers` files once per process. With per-session markers (`parsing_in_progress.<ulid>`), an abruptly terminated prior unit-test process can leave an unlocked active marker that the next test run recovers, pre-marking a language as failed.

## Why not delete the prefix files

Blindly removing every `parsing_in_progress.*` entry is unsafe because the asset directory can be shared by concurrently running Cargo processes. On Unix it can unlink a live peer`s locked marker, erasing that process`s crash evidence.

## Suggested direction

Keep parser/query assets shared, but give every unit-test process a unique crash-recovery state directory, matching the per-spawn `KAKEHASHI_STATE_DIR` isolation already used by the E2E LSP client helper. Avoid process-global environment mutation in parallel tests; inject or derive the test-only state path at registry construction.

## Evidence

`src/install.rs::test_data_dir()` explicitly shares `deps/` across runs and removes only the legacy marker. `tests/helpers/lsp_client.rs::isolated_state_dir()` already demonstrates the safe split between shared install assets and isolated crash state.

Found while revising #727; kept out of that production crash-recovery PR because it requires a broader test-path separation.

atusy:
## Summary

Deletes `FailedParserRegistry` and everything that fed it: the `parsing_in_progress`
marker, the `failed_parsers` blocklist, the three `is_parser_failed` gates on the
parse paths, `InstallOutcome::ParserFailed`, and the `KAKEHASHI_STATE_DIR` plumbing
that existed only to keep concurrent test servers from poisoning each other's crash
state.

Auto-install itself is untouched — only the crash-quarantine that shared
`AutoInstallManager` with it.

## Why

**It never detected the crash class it was built for.** `begin_parsing` only bumped
an in-memory `DashMap` counter (`origin/main:src/language/failed_parsers.rs:130`;
its own tests `test_begin_parsing_does_not_write_to_disk` /
`test_end_parsing_only_clears_memory` assert this). The `parsing_in_progress` file
was written solely by `persist_state`, whose only production callers were
`shutdown_impl` and the SIGTERM/SIGHUP watcher. A SIGABRT from a C assertion runs
neither, so startup found no marker and quarantined nothing — exactly what #725
reported. What it *could* fire on is a SIGTERM mid-parse, i.e. an editor restart:
that permanently disabled a perfectly healthy grammar, invisibly (`kakehashi
language status` scans only `parser/` and `queries/`, so a quarantined language
still printed `✓ parser ✓ queries` while being silently skipped).

**And the quarantine could not be undone.** It keyed on a bare language name in a
monotonically growing file. No production path ever removed an entry, so
reinstalling, upgrading, or repointing a parser left the language unparseable
forever. The warning it emitted advertised `kakehashi language clear-failed <lang>`,
a subcommand that was never implemented — the only real remedy was deleting
`<data-dir>/failed_parsers` by hand.

**Its one plausible trigger here was our own bug.** The only parser crash in this
project's history is #348: `parse_document` reused an unedited tree as the
incremental base, violating tree-sitter's contract and corrupting the markdown
scanner's heap. Had the marker worked, it would have quarantined `markdown` for
every affected user — and #349's fix would not have released them. A permanent
blocklist keyed on the language name punishes the grammar for a defect in the host.

**The failure it aimed at is already user-recoverable.** Set `autoInstall = false`
and delete the parser: `has_parser_available` goes false, `detect_language` returns
`None`, and the document degrades to a resolved-but-tree-less snapshot while every
other language keeps working.

**Cost avoided.** #727 proposed making the marker durable by restoring a synchronous
write on both sides of every host-document parse, each taking an exclusive
`crash_recovery.lock` shared by every kakehashi process using the same state dir — a
cross-process serialization point on the didChange path. Removing the mechanism
retires that whole line of work. (On `main` as it stands the parse-path cost was only
two DashMap shard ops per parse; the removal is not pitched as a perf win.)

Closes #725. Closes #726. Closes #730. Supersedes #727.

## Upgrade notes

- Any user currently carrying a **false-positive quarantine** from a
  SIGTERM-during-parse gets that language silently restored by this upgrade.
- `<data-dir>/failed_parsers` and `parsing_in_progress` are now inert. Nothing reads
  them and neither `language status` nor `uninstall --all` enumerates them, so no
  cleanup is required; they can be deleted.
- Caveat on the manual workaround: removing a parser also drops that language out of
  `detect_language`, and host bridging routes on a registered tree-sitter parser — so
  host-bridged requests and pushed diagnostics for that language stop too, even though
  `didOpen` still eager-opens the downstream server from the raw `languageId`. That is
  pre-existing coupling (the host-bridge parser gate), not introduced here, but it
  makes "delete the parser" a bigger hammer than tree-sitter features alone.

## Test changes

- `stale_install_task_stops_before_install_events` used `mark_failed` to force an
  early return. It now holds a test install claim, which makes a regressed staleness
  guard park on the `AlreadyInstalling` completion wait, and bounds the call in a
  timeout that turns that park into a fast, diagnosable failure. Verified empirically:
  with the guard at `install.rs:165-167` commented out the test fails in 5.02s naming
  the violated contract, versus 0.05s green at HEAD. The surviving `socket.next()`
  assertion independently covers the "guard moved below `try_install`" regression.
- Manager tests covering the registry are deleted; all of them exercised only removed
  code.

## Validation

- `cargo test --lib`: 3105 passed, 0 failed
- Full e2e sweep (`scripts/test_e2e_parallel.sh`): 57 binaries, 1696 tests, 0 failures
- `cargo check --all-targets --features e2e`, `cargo clippy --lib --features e2e -- -D warnings`, `cargo fmt --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience after parser crashes by removing persistent crash-state tracking and recovery marker persistence.
  * Automatic language installation now proceeds without being blocked by previous parser-failure history.
* **Documentation**
  * Clarified parse snapshot states (including settings-reload placeholders and “resolved-but-tree-less” outcomes).
  * Updated panic/log-target documentation to better describe behavior and states.

## Found during review, filed separately

- **#923** — `selectionRange` returns `null` during a settings reload instead of settling for the reparse. `Document::invalidate_parse` publishes a tree-less placeholder stamped at the bumped `content_version`, so the reader's currency check accepts it immediately. Pre-existing (this PR changes only comments in both files) and needs snapshot provenance to fix properly, so it is not folded in here; `selection_range.rs` now points at the issue.


atusy:
Fixed in 995b70b6a — thanks, that was a real miss. `test_data_dir`'s doc now says read-only-after-install is what makes the shared dir safe for parallel spawns, with no reference to the deleted `isolated_state_dir`.

Swept the two sibling helpers (`tests/e2e_cli_diagnose.rs`, `tests/e2e_cli_format.rs`): their `data_dir()` docs never mentioned the state dir, so no change needed there.

atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
Fixed in b375f0773. The nitpick was correct and pre-existing — this PR had only touched the neighbouring paragraph — but it stated a false invariant, so it is worth closing out here rather than leaving.

Clause 2 now matches `SnapshotSlot::admits` (src/document/snapshot.rs:158-163): strict `>` **and not a tree downgrade**, **or** equal version **and a tree upgrade**. The equal-version arm is load-bearing — it is how a reparse attaches its tree over a same-version tree-less publish (the settings-reload placeholder, the give-up snapshot). Documenting only `>` described a system that would strand those documents tree-less until the next edit.

PR #925, opened 2026-07-30
feat(config): make autoInstall per-language

atusy:
## Summary

`autoInstall` was global-only, so one crashing or unwanted grammar forced the whole
feature off. This adds `[languages.<lang>].autoInstall`:

```toml
[languages._]
autoInstall = true    # canonical spelling of the old global switch

[languages.python]
autoInstall = false   # never auto-install this one
```

The reverse works too — `_ = false` with per-language opt-ins. Precedence is the
language's own value → the `"_"` wildcard's → the deprecated top-level
`autoInstall` → `true`. Unset is the default at every level, so a language inherits
`_` and its `base` chain exactly like the other `LanguageSettings` fields.

Top-level `autoInstall` still answers when nothing per-language is set, and now gets
the same one-per-session migration notice `rootMarkers` has.

## The design constraint that shaped this

**`languages.*.autoInstall` outranks the top-level key regardless of which config
source supplied it.** That follows from the spec — the top-level key is only a
fallback for an unset per-language value — but it means key specificity can beat
*source* precedence, which the merge strategy otherwise guarantees. Two consequences,
both now enforced:

- `default_settings()` (merge layer 1) must not carry `languages._.autoInstall`. A
  built-in value there would silently shadow every user-supplied top-level opt-out.
  Its top-level value *is* safe — a user overlays the same key and wins.
- The `config init` template must emit **neither** spelling. The top-level one would
  make kakehashi warn about a file it wrote itself; `[languages._]` would let a
  generated user config (layer 2) outrank an `autoInstall = false` pushed via
  `initializationOptions` (layer 4). `autoInstall` is therefore the documented
  exception to config-init-template-convention: the default lives in the resolver and
  the docs, not the generated file.

The second point was a regression in the first version of this branch, caught by
review and fixed in `be826d174`. It is now pinned by
`a_generated_template_does_not_shadow_a_higher_layer_optout` and, through the real
binary, `test_config_init_emits_no_auto_install_spelling`. `docs/README.md` and the
merge-strategy ADR both record the precedence exception.

## Other things reviewers should know

- **The deprecation detector checks only the top level.** `[languages.*]` carries the
  canonical key of the same name, so a name-only scan would warn about exactly the
  config the notice tells people to write. Pinned by
  `toml_does_not_flag_the_canonical_per_language_auto_install` and the e2e
  `..._spares_the_canonical_key`.
- **A present language entry owns the answer.** `resolve_base_configs` folds `"_"` and
  the `base` chain into every present entry, so `auto_install_decision` consults `"_"`
  only for languages the fold never ran for. Otherwise a language that escaped
  wildcard inheritance with a self-referential `base` would still pick `"_"` up for
  this one field.
- **The disabled-reason message deliberately hedges on the language arm.** After the
  fold, a `Some(false)` on the entry could have come from the language, its base, or
  `"_"`, so naming `languages.<lang>.autoInstall` as the culprit would misdirect
  someone hunting the global switch. It names the overriding key and the possible
  sources; the wildcard and top-level arms stay exact.
- **`injection.rs` hoisted the gate above its per-language loop.** It now resolves
  inside, keyed on `resolved_lang` (the language actually installed), so one injected
  language's opt-out cannot decide for its siblings.
- `rename_all = "camelCase"` on `LanguageSettings` renames nothing — every
  pre-existing field is one word — but is load-bearing now, and
  `known_language_setting_keys_match_schema_properties` pins it.
- The Neovim quickstart snippets set `autoInstall` in kakehashi's own `init_options`
  (merge layer 4), so they would have popped the new warning every session. Updated to
  the canonical key.
- One pre-existing test needed adjusting:
  `did_change_configuration_merges_into_settings_published_while_waiting` pushes
  top-level `autoInstall` on purpose and never drains its socket, so the new
  `show_warning` parked the future for a reason unrelated to the lock it is meant to
  park on. It consumes the notice slot up front.

## Test rigor

Review ran mutation experiments and found three one-line edits that left the whole
suite green. All three now fail:

| Mutation | Now caught by |
|---|---|
| drop `merge_language_settings`'s `auto_install` arm | `auto_install_inherits_through_the_base_chain` (uses a `base` chain, not `_`, so it actually discriminates) |
| `DeprecatedKeysSeen::merge` overwrites instead of OR-ing | `merge_ors_flags_so_a_clean_later_layer_cannot_clear_them` — an empty `initializationOptions` runs every session and would have cleared a file-detected flag, so the notice would never fire |
| `run_config_init` → `default_settings()` | `test_config_init_emits_no_auto_install_spelling` |

Verified by re-applying each mutation in a scratch worktree.

## Validation

- `cargo test --lib`: 3128 passed, 0 failed
- Full e2e sweep: 57 binaries, 1698 tests, 0 failures
- `cargo clippy --lib --features e2e -- -D warnings`, `cargo fmt --check`

## Summary by CodeRabbit

* **New Features**
  * Added per-language `autoInstall` control with wildcard defaults, including language-specific behavior for missing parsers/queries.
  * When auto-install is disabled, users now get clearer, language-specific disable explanations.
  * Added session-scoped deprecation warnings for the legacy top-level `autoInstall` setting.

* **Documentation**
  * Updated migration guidance for moving from top-level `autoInstall` to `languages._.autoInstall`.
  * Refreshed Neovim and `config init` examples to avoid generating deprecated `autoInstall` keys.

* **Bug Fixes**
  * Fixed auto-install evaluation for injected languages to be independent.
  * Corrected precedence and inheritance behavior, including removing redundant inherited overrides and preserving user opt-outs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,6 @@ To let kakehashi auto-install the parsers it needs, make sure a **C compiler** i
 ```lua
 vim.lsp.config.kakehashi = {
     cmd = { "kakehashi" },
-    init_options = { autoInstall = true },
     on_attach = function(_, bufnr)
         -- Let kakehashi own highlighting (avoids double-highlighting)
         vim.api.nvim_create_autocmd("LspTokenUpdate", {
--- a/docs/README.md
+++ b/docs/README.md
@@ -240,7 +240,7 @@ Per-language configuration. Usually not needed as kakehashi auto-detects languag
 
 | Field | Description |
 |-------|-------------|
-| `base` | Inherit parser, queries, and bridge configuration from another language |
+| `base` | Inherit parser, queries, bridge, and `autoInstall` configuration from another language |
 | `parser` | Explicit path to the parser library (`.so`, `.dylib`, `.dll`) |
 | `queries` | Array of query configurations with `path` and `kind` (highlights, bindings, injections) |
 | `bridge` | Per-injection-language bridge filter and aggregation settings |
@@ -252,10 +252,19 @@ Per-language configuration. Usually not needed as kakehashi auto-detects languag
 Whether kakehashi may download and install a missing parser/queries for this
 language when a file is opened.
 
-Resolved most-specific-wins: the language's own value, then the `"_"` wildcard's,
-then the deprecated top-level `autoInstall`, defaulting to `true`. Unset is the
-default at every level, so a language inherits `"_"` exactly like the other
-fields.
+Resolved most-specific-wins: the language's own value, then each entry in its
+`base` chain, then the `"_"` wildcard's, then the deprecated top-level
+`autoInstall`, defaulting to `true`. Unset is the default at every level, so a
+language inherits through `base` and `"_"` exactly like the other fields — for
+example `[languages.rmd] base = "markdown"` picks up markdown's `autoInstall`
+before `"_"` is consulted.
+
+**Which name to use.** The key names the language kakehashi would *install*, not
+the token in your document. A host document uses the `languageId` your editor
+sends; an injected region uses its resolved language, so a ```` ```py ```` fence
+is governed by `languages.python.autoInstall` and an `rmd` region whose own
+parser is absent is governed by whatever its `base` resolves to. Set the key on
+the resolved name.
 
 Enable everywhere except one language:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -154,8 +154,10 @@ default `info` forwards Error, Warning, and Info while suppressing LSP `Log` and
 ```json
 {
   "searchPaths": ["$HOME/.local/share/kakehashi", "/another/path"],
-  "autoInstall": true,
   "languages": {
+    "_": {
+      "autoInstall": true
+    },
     "typescript": {
       "base": "ecma",
       "queries": [
@@ -222,10 +224,15 @@ Array of base directories to search for parsers and queries. If not specified, u
 Parsers are searched as `{searchPath}/parser/{language}.{so,dylib,dll}`.
 Queries are searched as `{searchPath}/queries/{language}/{query_type}.scm`.
 
-#### `autoInstall`
+#### `autoInstall` (deprecated)
 
-- `true` (default): Automatically download and install missing parsers/queries when a file is opened
-- `false`: Require manual installation via CLI
+**Deprecated:** use [`languages[*].autoInstall`](#languagesautoinstall) instead.
+The top-level key still works — it answers whenever no per-language value is
+set — but kakehashi shows a one-time migration notice when it is present, and
+it may be removed in a future release.
+
+Move `autoInstall = true` to `[languages._] autoInstall = true` for identical
+behavior, then override per language as needed.
 
 #### `languages`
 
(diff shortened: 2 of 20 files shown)


atusy:
Bot round 1 — responses, newest commit `300494a18`.

**Copilot, `deprecation.rs` (TOML syntax in a notice shown to JSON users) — valid, fixed in `300494a18`.** The notice now uses dotted paths (`languages._.autoInstall`), with a comment recording why: it also fires for `initializationOptions` and `didChangeConfiguration`, where `[languages._]` names a shape the user cannot write.

**Codex + Qodo, "wrong autoInstall key reported" / "preserve the inherited source" — already fixed in `be826d174`, which predates your review of `04eda1ff3`.** You are right about the cause: `resolve_base_configs` flattens `_` and the `base` chain into every present entry, so a folded `Some(false)` cannot be attributed to the language. The message no longer claims it — the language arm now reads ````languages.python.autoInstall` resolves to false (set on the language, its `base` chain, or `languages._`)```, and the wildcard/top-level arms stay exact because those are reliable. Pinned by an exact-string assertion in `auto_install_disabled_reason_points_at_the_config`.

**Codex, "respect self-referential bases before applying the wildcard" — also already fixed in `be826d174`.** `auto_install_decision` no longer re-consults `_` for a language that has an entry; `_` is read only when the fold never ran for that language. Pinned by `auto_install_respects_a_self_referential_base_through_the_real_fold`. The migration consequence you'd expect from that — moving the top-level key to `languages._` is not equivalence-preserving for a self-referential `base` — is now called out in both the notice text and `docs/README.md`.

Full sweep after these: `cargo test --lib` 3128 passed, e2e 57 binaries / 1699 tests / 0 failures.

atusy:
Pipeline converged. Summary of the review rounds, since the branch changed substantially:

**Two real defects were found and fixed, both mine:**

1. **The `config init` template inverted layer precedence** (`48cadf49b`). It wrote `[languages._] autoInstall = true`, and since `auto_install_for` checks `_` before the top-level key and is layer-unaware, a generated user config (layer 2) silently outranked an `autoInstall = false` pushed at layer 4 — failing in the direction that hurts. Reproduced by three independent probes. The template now emits neither spelling; `autoInstall` is the documented exception to config-init-template-convention, because neither spelling can be generated inertly.
2. **The disabled-reason message shipped with 18 literal spaces** mid-sentence (`76ef533a0`) — a lost line-continuation, invisible to a `contains` assertion. Now pinned by an exact-string comparison.

**Two semantic corrections:** a present language entry now owns the answer (it has already been through the base-chain fold, so re-consulting `_` overrode the deliberate self-referential-`base` escape), and the reason message no longer claims the language key turned it off when the fold makes that unknowable.

**Six mutation gaps closed**, each verified by re-applying the mutation in a scratch worktree — dropping the merge arm, reversing it, `DeprecatedKeysSeen::merge` overwriting instead of OR-ing, repointing `run_config_init`, deleting the initialize `show_warning`, and a duplicate notice for one update. Every one of those previously left the whole suite green.

**One bug in a doc example:** the Neovim snippet ended up with `languages` twice in one Lua table, so the second silently shadowed the first and nullified the line it was added for.

**Documented rather than changed**, with reasons in code and docs: a hand-written `languages._` in a low-precedence source outranking a higher-precedence top-level value (a fix needs per-key provenance the config types don't carry); `didChangeConfiguration` merging onto the previous snapshot so unsetting can't restore a default (pre-existing for every Option field); a `base` chain terminating before `_` inheriting nothing from it (migration caveat, now stated in the notice text too); and host vs injected paths keying the gate on different name forms, each matching what that path installs.

One unrelated flake surfaced during a sweep — `e2e_whole_document_link_cancel_forwards_to_concatenated_layers` in `e2e_host_bridge`, a cancel-timing test. 1 failure in ~8 sweeps on this branch, 4/4 clean on `main`, and two clean sweeps after rebasing. No mechanism connects it to this change (the gate's disabled path only runs for a language whose parser failed to load). Flagging rather than burying it.

**Final state:** `cargo test --lib` 3129 passed; two consecutive full e2e sweeps at 57 binaries / 1699 tests / 0 failures; clippy and fmt clean; rebased onto current `main`.

PR #926, opened 2026-07-30
feat(bridge): route executeCommand by connection, not by document

atusy:
## Why

`workspace/executeCommand` carries no `textDocument`. An ordinary language server needs none — it already knows its own workspace and its own open documents, so a command name is a complete instruction. That knowledge is not absent, it is *implicit*.

Kakehashi has no such implicit knowledge: it multiplexes N downstream servers across M workspace roots, so "which process runs this command" must be carried explicitly, and the command **name** is the only channel (this method has no `data` envelope).

The name carried a host **document** — more than routing needs. It was there to serve a second, unrelated purpose: repairing virtual documents a downstream respawn purged, because unlike the request paths this one has no `ensure_document_opened` step. Conflating the two is what made the command set per-document, hence unbounded, hence unadvertisable — the reason clients that dispatch only registered command ids show a bridged action that never runs.

## What

Split the two concerns.

**Routing token names a connection:**

```
kakehashi|{root_tag}|{server}|{root}|{command}
```

`root_tag` distinguishes the three rooting modes (marker / client-root fallback / shared instance) — collapsing any pair would run the command in the wrong workspace. Routing rebuilds that `ConnectionKey` and reaches the connection directly; acquiring with `document_uri: None` would resolve to the client-root fallback and replace a marker-rooted process with a client-rooted one.

Escaping (`%` first, then the separator) makes the split boundaries separator-free **by construction**. The old scheme could only *enforce* that by refusing to mint a name when a config key contained the separator, so this removes the unroutable-command state entirely — along with `REASON_UNROUTABLE_COMMAND` and both `command_dropped` disable paths, which existed only to surface it.

**Repair moved to respawn.** The purge is the last moment the affected set is knowable — it is the purge itself that forgets what the dead process held — so `purge_connection` returns the host documents it dropped, and they are re-opened when the replacement reaches Ready. This covers every request path, not just `executeCommand`.

Repair being asynchronous costs the ordering the inline `await` gave for free. Two things restore it: the re-open **awaits** the per-server open (the eager batch path only spawns, so completing it proves nothing), and the pending re-open is claimed **before** the connection is published as Ready so a request unblocked by that transition can see it and wait under the old 2 s bound.

## Deliberately deferred

Dynamic registration of the encoded names. Making the set finite was the precondition, not the feature; registering them needs two answers only a real client can give (tolerance of repeated overlapping `workspace/executeCommand` registrations as roots are discovered lazily, and whether machine-generated ids are acceptable). `executeCommandProvider.commands` stays empty and the vscode-languageclient limitation stands. Recorded in the ADR's gap section.

A **dead** shared-instance connection still cannot be revived from the token alone — it is keyed without a root, and announcing its workspace folders needs a marker only a document can produce. A live one, including one mid-handshake, is routable.

## Notes for review

- `docs/architecture-decisions/execute-command-routing-token.md` carries the reasoning, the rejected alternatives (opaque handle + mint table; recovering the host from `arguments`), and the gaps.
- Old-format names fail closed: the new decoder rejects them at `strip_prefix('|')` before constructing any key, so a client that cached an action across an upgrade gets one fail-soft null. Names are minted per codeAction response and never persisted.
- Two commits are review fixes; their messages name the finding.

Gate: `cargo clippy --all-targets --all-features` clean, 3148 lib tests, 114 e2e (`e2e_code_action` / `e2e_host_bridge` / `e2e_organize_imports`).

## Summary by CodeRabbit

* **New Features**
  * Improved `workspace/executeCommand` routing using connection-based command tokens (including palette-command prioritization).
  * Added coordinated “re-open virtual documents” handling during downstream reconnection, with a completion signal for correct sequencing.
  * Updated bridged code-action command routing to use the same connection-based token envelope.
* **Bug Fixes**
  * Prevented out-of-order command delivery by gating dispatch on pending re-opens; timeouts now fail softly instead of stalling.
* **Documentation**
  * Updated LSP Bridge and architecture decision docs to clarify routing token behavior, trust boundaries, and reconnect ordering/limits.
* **Tests**
  * Expanded coverage for the new routed-command token format and verified reopen ordering across respawns.
* **Chores**
  * Updated Tokio feature flags.


atusy:
## Review round 1 responses

All findings addressed. Fixes are one commit each, referencing the finding.

| Finding | Source | Outcome |
|---|---|---|
| Initializing branch returns handle without rechecking launch config | codex-connector, Qodo, codex | Fixed `504efd321` — recheck **after** `wait_for_ready`; the config is only recorded during the handshake, so checking before it was vacuous |
| Barrier signals `done` on budget expiry, so the command sends unordered anyway | codex | Fixed `dba7455b9` — `wait_for_reopen` now reports whether it settled and both routes **fail soft** rather than send. A dropped sender still counts settled (that re-open can never finish) |
| Raw palette command shaped like a routed name gets mis-decoded | codex | Fixed `dba7455b9` — route the palette registry first; it holds raw names only, so an exact hit is unambiguous |
| `decode_command` accepts shapes `encode_command` cannot emit (stray `%`, root-less tag with a root) | CodeRabbit | Fixed `504efd321` — fails closed, falls through to palette handling |
| Palette path lacked the re-open barrier | codex | Fixed `ab0e3a3d7` |
| tokio `sync` feature undeclared | CodeRabbit | Fixed `d6d24fb4b` — arrived via feature unification; now explicit |

### Refuted

**An invented server name is spawnable via a `_` wildcard** (codex). `is_server_spawnable` (`src/config/merge.rs:48-58`) does `servers.get(name).is_some_and(...)` — an absent name short-circuits `false` before the wildcard is consulted. The wildcard supplies defaults to an existing entry; it never creates one. `dispatch_execute_command` calls this before `resolve_with_wildcard`, so `kakehashi|m|invented|file:///etc|cmd` is dropped with a warn.

The *other half* of that finding — a configured server reachable at an arbitrary root — is real and accepted rather than fixed, with the premise it rests on recorded at the top of `command_routing.rs` and in the ADR: the client that could forge a token is the client that supplies `languageServers` (each with its `cmd`) via `initializationOptions`, so it already decides what runs and where.

### Known, not fixed here

`ReopenDocuments` carries `server`, not the claimed `ConnectionKey`, so if `workspaceMarkers` changes between purge and respawn the re-open can resolve a different key than the token names. Largely mitigated by the launch-config recheck above (a `workspaceMarkers` change *is* a launch-config change, so the stale connection is now rejected), but not structurally closed — that needs a key-targeted open. Tracking separately rather than growing this PR.

**Still wanted before merge:** an e2e that kills the mock downstream mid-session and asserts `didOpen` precedes `workspace/executeCommand` in its message log. The barrier-releases-early bug appeared four times across review; every unit test drives the completion signal by hand and structurally cannot catch a fifth.

Gate on `d6d24fb4b`: clippy `--all-targets --all-features` clean, 3150 lib tests, 114 e2e.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
## Round 2 responses (`1d328585f`)

**Fixed — barrier must stay pending until the background work finishes.** Correct, and it is the *fifth* instance of one bug on this branch. `dba7455b9` made the **waiter** report failure on its timeout but left the handler signalling `true` on **its** timeout, so an exceeded budget still marked the re-open complete while its `didOpen`s were queued.

`done` now moves into the work task, so only real completion signals it. The handler's timeout stops *watching* without claiming anything: requests still waiting time out and fail soft, a later request waits on the same barrier, and a panic drops the sender (read as "can never finish"). "Bound the wait, never the work" now holds for the signal too, which is what I got wrong.

**Not fixed — re-open resolves against current settings rather than the claimed key.** The reasoning is now a comment at the call site so it does not get re-raised.

`workspace_markers` is one of the fields `same_launch_config` compares (`src/lsp/bridge/pool.rs:64-90`). So after a markers change, **both** by-key routing paths reject the old connection — `ready_connection_by_key_for_config` on the fast path, and the post-handshake recheck added in `504efd321` on the Initializing path. The command therefore fails soft rather than executing document-less, which is the outcome the finding is concerned with. Opening the document at the newly resolved location is correct for the document itself.

Carrying the claimed `ConnectionKey` would let the re-open target the dead key exactly, but that buys nothing while the key is unreachable by construction. Happy to reconsider if there is a path to the old key I have missed.

Gate: clippy `--all-targets --all-features` clean, 3150 lib tests, 114 e2e.

atusy:
## Round 3 — no new findings

Both comments on `1d328585f` are carry-forwards of findings already closed. Refuting each once, per the review convention, and resolving the threads.

**Qodo, "Stale config bypassed on init" (`pool.rs:1275`)** — already fixed in `504efd321`, before this commit. `reconnect_by_key` now rechecks the launch config *after* `wait_for_ready` (`src/lsp/bridge/pool.rs:1256-1266`):

```rust
Ok(()) => {
    if handle.launch_config().is_some_and(|live| !same_launch_config(live, config)) {
        log::warn!(target: "kakehashi::bridge",
            "executeCommand: connection {key} finished initializing under a superseded config; ignoring");
        return None;
    }
    Some(handle)
}
```

The recheck has to be *after* the handshake, not before it — `launch_config` is only recorded during the handshake, so a pre-check would read `None` and pass vacuously. The comment is anchored to the pre-fix revision.

**codex-connector, "Reopen documents on the connection recorded by the purge" (`lifecycle.rs:1317`)** — answered in the round-2 comment above and now documented at the call site. Short form: `workspace_markers` is one of the fields `same_launch_config` compares (`pool.rs:64-90`), so after a markers change both by-key routing paths reject the old connection and the command fails soft rather than executing document-less. Opening the document at its newly resolved location is correct for the document. Carrying the claimed key would target the dead key exactly, which buys nothing while that key is unreachable by construction.

---

### `/revise` complete

| Stage | Rounds | Result |
|---|---|---|
| `/review`, 10 perspectives | 1 | 6 findings fixed |
| codex MCP (thread, then fresh session) | 2 | converged; 4 fresh-session findings fixed |
| PR | — | this one |
| PR bots | 3 | round 3 = zero new actionable comments |

**Merge blocker I'd keep open:** the ordering e2e — kill the mock downstream mid-session, assert `didOpen` precedes `workspace/executeCommand` in its message log. A barrier signal outrunning its work was found **five** times across these rounds (fire-and-forget; registered after Ready; awaiting a function that only spawns; signalling on the waiter's timeout; signalling on the handler's timeout). Every unit test drives the completion signal by hand and structurally cannot catch a sixth.

Gate: clippy `--all-targets --all-features` clean, 3150 lib tests, 114 e2e.

atusy:
## Merge blocker resolved (`de8eba75d`)

The ordering e2e now exists and is **verified discriminating by mutation**:

- `done.send(true)` moved before the opens → test FAILS
- waiter ignoring the unsettled barrier → test FAILS
- unmutated → passes; full `e2e_code_action` suite 44/44, twice

Design: an e2e-only stall (`KAKEHASHI_E2E_STALL_REOPEN_MS`, gated on the existing `e2e` cargo feature — release builds do not contain the branch) holds the respawn re-open past `REOPEN_WAIT`, so a correct kakehashi has exactly one legal outcome: drop the command fail-soft. Any signal-outruns-work shape sends it ahead of the `didOpen`, caught by wire-order in the segment after the replacement's `initialize`. The mock gained an append-only wire log shared across incarnations and an atomic crash-once marker.

A first attempt without the stall was built and discarded: reintroducing the bug still passed, because locally the re-open wins the race by accident. Racing proves nothing; the stall makes the barrier load-bearing.

Gate: featureless clippy clean (proves the hook is absent from release), `--all-targets --all-features` clean, 3150 lib tests, e2e green.

PR #927, opened 2026-07-30
fix(bridge): re-open the connection the purge claimed

atusy:
Follow-up to #926. Raised there three rounds running and each time refuted as harmless rather than closed; this closes it.

## The problem

When a downstream connection is purged, the host documents it held are re-opened on its replacement, and a barrier makes a routed `workspace/executeCommand` wait so it cannot overtake its own `didOpen`.

`ReopenDocuments` carried only the server name, and the open resolved its connection from the **host document**. So a `workspaceMarkers` change between purge and respawn re-rooted the host, the opens landed on a newly resolved connection, and the claimed one — the one the barrier signals for, and the one a routed token names — stayed empty.

The old refutation was that no command could reach the claimed connection anyway, because `workspace_markers` is compared by `same_launch_config`. **That is wrong**, and this PR corrects the comment that stated it: a routing token carries the root, `reconnect_by_key` rebuilds the workspace from the token rather than from current markers, and the connection it spawns records the *current* config — so no launch-config check rejects it. The root is not a config field at all, so config comparison cannot see a stale root.

## The fix

- `ReopenDocuments` carries the claimed `ConnectionKey` (and no longer the server separately — two sources for one fact is an unchecked invariant).
- The open is **acquired by that key**, not merely checked against it. Ready-only and never spawning: it repairs a connection that exists, and a named connection that has since died has nothing to restore.
- The barrier reports the truth. `eager_open_virtual_documents` returns an `OpenOutcome`, the re-open tracks whether the claimed connection actually got its documents, and `done` carries that. `wait_for_reopen` distinguishes "repaired" from "the sender went away without reporting success", so a command is withheld rather than released onto an empty connection — while still retiring the entry, since nothing will ever settle it.

## Review note

The middle commit here did **not** do what its subject claimed — it pinned the check but not the acquisition, turning "repairs the wrong connection" into "repairs nothing". Caught in review; the last commit fixes it and the message says so. Left in history rather than squashed because the distinction is the interesting part.

Gate: `cargo clippy -- -D warnings` (the CI invocation) clean, `--all-targets --all-features` clean, 3168 lib tests, e2e 44/44 twice.

### DIFF
--- a/docs/architecture-decisions/execute-command-routing-token.md
+++ b/docs/architecture-decisions/execute-command-routing-token.md
@@ -96,6 +96,14 @@ injection regions can supply, so the work is performed on the server side
 the existing pool→editor upward request channel. The pool decides *when*; the
 server side decides *what*.
 
+The pool also states *which connection*. The server side resolves each host
+against current settings, so a config change that re-roots the host between the
+purge and the respawn would otherwise repair a different connection than the one
+the barrier signals for — leaving the claimed one empty while reporting success.
+The claimed key travels with the request and the open is pinned to it; a
+re-routed host is skipped rather than opened elsewhere, since its new connection
+opens it through its own request path.
+
 That split makes the repair **asynchronous**, where the inline version was
 awaited — and being awaited was load-bearing, not incidental. The outbound
 queue to a downstream is FIFO, so a request enqueued before the re-open's
--- a/docs/architecture-decisions/execute-command-routing-token.md
+++ b/docs/architecture-decisions/execute-command-routing-token.md
@@ -100,9 +100,17 @@ The pool also states *which connection*. The server side resolves each host
 against current settings, so a config change that re-roots the host between the
 purge and the respawn would otherwise repair a different connection than the one
 the barrier signals for — leaving the claimed one empty while reporting success.
-The claimed key travels with the request and the open is pinned to it; a
-re-routed host is skipped rather than opened elsewhere, since its new connection
-opens it through its own request path.
+The claimed key travels with the request and the open is ACQUIRED by it, not
+merely checked against it — resolving from the host would repair whichever
+connection it routes to now. A claimed connection that has since died is
+reported as unrepaired, and the barrier then withholds the commands waiting on
+it rather than releasing them onto an empty connection.
+
+A stale-rooted claimed connection is reachable rather than hypothetical: a
+routing token carries the root, and reconnection rebuilds the workspace from the
+token rather than from current markers, so it spawns a connection recording the
+current config that no launch-config check can reject. The root is not part of
+the server config, so config comparison cannot detect a stale root at all.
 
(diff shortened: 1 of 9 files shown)


atusy:
## Open findings from codex review — not fixed, deliberately

Two survive on this HEAD. Both are real; neither is a regression against `origin/main`, and I am recording rather than rushing them because the second is a design question, not a bug fix.

### 1. Ready-only acquisition can miss an `Initializing` replacement (`did_open.rs:92`)

Interleaving: connection A reaches Ready and queues `ReopenDocuments`; before the handler runs, A dies and another acquire installs B under the same key as `Initializing`. The Ready-only lookup returns `NotOpened` even though B is about to be Ready.

Consequence is fail-soft — the command is withheld and the user re-fires — so it errs in the safe direction. Fix would be a bounded wait on an existing matching `Initializing` handle, then re-check config and pool identity (the same shape `reconnect_by_key` already uses).

### 2. Batch granularity of `done` (`lifecycle.rs:1406`)

`repaired` is false if **any** host failed, so one unrelated host withholds commands the successfully-opened hosts would have served. And the guarantee does not persist: once the sender drops, the barrier retires, so a later command for the still-unrepaired host proceeds anyway.

So the current signal is neither precise nor durable. The honest framing: the barrier is per-**connection** while repair succeeds or fails per-**host**, and a routed token names a connection, not a host — so nothing lets us tell which host a given command needs. Conservative-withhold is defensible for correctness, but choosing between that and per-host tracking (which needs commands to carry a host, i.e. re-introducing what #926 removed) is a decision worth making deliberately rather than at the end of a long session.

I would rather leave this stated than half-change the barrier semantics — the failure mode throughout #926 was exactly that: fixing one side of a signal and leaving its mirror.

Everything else codex checked is clean (the three-way retire policy in `wait_for_reopen`, generation-checked retirement, no permanently-blocking entry).

atusy:
## All three open threads fixed — `11af62e42`

Every one was the same omission: reporting `Opened` where the claimed connection did **not** end up with its documents. A barrier is only worth having if its success signal is honest, so these are squarely in scope.

| Thread | Fix |
|---|---|
| codex-connector, `did_open.rs` — failed `didOpen` enqueue reported as a successful repair | Track it; keep opening the rest (a partial repair still helps) but return `NotOpened` |
| codex-connector, `coordinator.rs` — removed server mapping reported as success | With no named connection that is genuinely nothing to do; with one, the host can no longer supply what was asked for, so the repair did not happen |
| Qodo, `pending_reopen.rs` — `false` documented as timeout-only | It has three meanings now; documented all three and the one consequence they share |

Gate: `cargo clippy -- -D warnings` clean, `--all-targets --all-features` clean, 3168 lib tests.

The two findings in my earlier comment (Ready-only vs `Initializing`; per-batch `done` granularity) remain open and unfixed, deliberately — reasoning there.

PR #931, opened 2026-07-31
fix(bridge): fence and refresh diagnostics on workspace folder changes

atusy:
## Summary

`workspace/didChangeWorkspaceFolders` replaces the project a downstream server analyses. What it reports about a file can change without the file changing — import resolution, project membership, "not part of the project" verdicts. Two things were missing on that path.

**1. The pull-diagnostic lineage was not fenced.** `propagate_settings` already refuses to let a `previousResultId` cross a settings replacement, and says why:

> A downstream may change diagnostic semantics without changing its document contents, so no previousResultId lineage survives a settings replacement safely.

A folder change is that event with a different input. `apply_workspace_folder_change` now calls `invalidate_diagnostic_connections` for every client-fallback key.

The case with no other guard is a connection that **takes** the notification and keeps running: it moves no generation at all — not settings, not document, not the connection generation a purge would bump — so a pull issued before the change and answered against the previous folder set passed the lineage gate, was published as current, and its `resultId` became the baseline the next request sent back. (A recycled connection was already partly covered, since `purge_connection` bumps the connection generation.)

**2. Nothing asked the editor to re-pull.** This is why the fence could not ship alone: a stale-lineage `Unchanged` resolves to `Vec::new()`, so fencing without a nudge can *empty* a diagnostic layer — strictly worse than the stale-but-populated state before. No refresh was requested anywhere on this path; the settings reload the handler performs asks only for a semantic-tokens refresh.

The handler now requests a forced `workspace/diagnostic/refresh` **after** the settings reload, and separately on the `--config-file` branch, which returns before any reload and is its only chance. Order matters: the editor answers a refresh by pulling immediately, and a pull served from the previous project's settings is not merely stale — if the new project drops the server that produced a layer, the reparse yields `PullLayerOutcome::Skip`, which neither clears the layer nor asks again, leaving the editor holding diagnostics from a server the new workspace does not configure.

Forced is load-bearing: the coverage gate suppresses an unforced refresh when nothing is dirty *by version*, which is exactly the shape of a change that moves the project rather than the documents.

## Scope

**Fenced**: `is_client_fallback()` keys, whose workspace *is* the client's by construction (`workspace_from_marker` falls back to the pool's root/folders when no marker root resolves).

**Exempt**: marker-rooted and shared connections, whose root comes from an on-disk marker walk that an upstream folder change does not touch. A `Shared` key is only ever minted for a marker-rooted acquisition, so it never carries the client snapshot.

**Ignored whole**: an event naming neither an addition nor a removal — no fence, no forwarded notification, no recycling, no reload, no refresh. `apply_workspace_folder_change` reports whether the event described a change, so one definition drives all four.

The snapshot mutation happens under the `connections` guard rather than before it, so two concurrently-dispatched folder events cannot apply to the snapshot in one order and reach the forwarding loop in the other.

## Review

Four rounds of local multi-perspective review plus two codex passes. Beyond the original change, they caught:

- The missing refresh, without which the change was net-negative for pull-namespace editors.
- The refresh being sent *before* the reload — my own comment had called that benign; the `PullLayerOutcome::Skip` path above is why it isn't.
- Guarding only the fence for empty events **introduced** a hole: the recycle loop is not conditional on the event saying anything, and the respawn path's own invalidation is gated on finding an existing connection the recycle has already removed — so an empty event could kill an incapable downstream and hand its `resultId` to the replacement. Fixed by ignoring such events entirely.
- The handler running the whole settings reload for an empty event, re-deriving the settings root (which drops the project config layer when the folder list is empty).
- The original ordering rationale was wrong: what closes the in-flight window is that a pull enqueues its request while holding `connections` (`execute.rs` for a virtual region, `host.rs` for a host document), not the statement's position relative to the notification.
- An e2e negative assertion that could not prove what it claimed — tower-lsp dispatches handlers concurrently, so the `shutdown` I used as a FIFO barrier can be handled before the notification ahead of it, and `shutdown` suppresses an already-admitted refresh. Replaced with an in-process assertion on a counter recorded ahead of every admission gate.
- Test fixtures that would have silently degraded to the already-covered recycle shape.

## Also in this branch, and not the subject

`tests/e2e_code_action.rs` had a setup wait that polled the mock's wire log for 15s waiting on an eager-open fan-out that nothing drives. It is now driven by a request — `ensure_document_opened` runs before the request is queued on the same connection FIFO, so a response carrying actions is proof the didOpen arrived. `tests/bin/mock_formatter.rs` gates its deliberate crash off the second host so that request does not kill the process under test. This is test hygiene; it is here because it was blocking a clean read of the suite during review.

## Follow-ups filed

- **#932** — `announce_shared_root` has this defect for a different trigger (a marker root joining a shared connection). On the per-request acquisition path, so an unconditional bump would be a hot-path regression; needs a lock-ordering decision of its own.
- **#937** — a fenced response still *reaches* the editor: a stale `Full` is returned verbatim, a stale `Unchanged` as empty, and the repair is not ordered against it. Applies to all three fences; the settings path, with no repair at all, is the worst.
- **#945** — the forwarded notification is queued before the connection's own folder set is updated, so a downstream can be told folders changed and then handed the pre-change list.
- **#938** / **#946** — two e2e tests whose remaining waits cannot be converted without new machinery, with designs for both.
- **#965** — the fence only covers the *pull*-diagnostic lineage; a still-alive client-fallback connection's cached *push* diagnostics (folded in by `fold_push_fallback_diagnostics`) are not evicted by a folder change, so a push-only server's stale diagnostics can survive it. Same "needs its own lock-ordering/keying decision" shape as #932.
- **#966** — no e2e test proves the refresh-after-reload ordering (`9f78ae0cc`) end-to-end: both refresh-focused e2e tests use `--config-file`, which never reaches the reload path the ordering fix is actually about. Needs new mock-server machinery to give two workspace folders visibly different server sets, same as #938/#946.

## Round 2 (post-rebase /revise, 2026-08-09)

Rebasing this far past main required a fresh look, so a full second review pass ran: 10 parallel independent local perspectives against the rebased code, then a fresh codex pass with no thread carryover from the pre-rebase rounds.

**Local pass** — 6 of 10 perspectives found nothing; 4 real, minor findings, all fixed:
- A doc comment on `apply_workspace_folder_change` still described the pre-rebase lock order (mutation "before" the guard); the rebase moved it under the guard and the inline comment beside it already said so, but the function doc didn't.
- `request_pull_diagnostic_refresh`'s doc comment enumerates every origin allowed to force past the coverage gate; it never gained this PR's new caller.
- The empty-event unit test only asserted the refresh counter, not lock ordering — added a test that races the handler against an externally-held `reload` guard (verified RED against a version with the acquisition moved after the pool call).
- No test verified the `--config-file` branch's `set_root_path` call independently of the refresh it sends alongside it — added an e2e test using the forced refresh request as a barrier, then a relative-path push to prove the anchor moved (verified RED against a version with the call dropped).

**Codex pass** — 4 findings:
- Two were real design gaps, filed as #965 and #966 above rather than folded in, matching how #932/#937/#945/#938/#946 were already scoped out of this diff for needing their own decisions or new test machinery.
- One (stale pull responses not ordered against the repair) is exactly #937, already filed.
- One was a real bug in the lock-ordering test added in this same round: the loop that polls for the pool mutation returned on *either* the mutation or the handler completing, so a regression that let the handler finish without ever touching the pool would still resolve the timeout as `Ok`. Fixed to return a `bool` distinguishing the two, and re-verified RED against the same class of regression (the pool call returning `false` unconditionally).

## Rebased onto main (2026-08-09)

Main gained #746/#954 in the interim, which rewrote `did_change_workspace_folders_impl` around this branch's changes: a `reload`-before-`connections` lock-ordering invariant (`a6c3d63de`), a root-derivation ladder that survives an emptied folder list via `config_root_after_folder_change` instead of this branch's original first-folder-only read, and publishing the new root only once the settings it derived are the ones in effect (`acc53819c`). None of that conflicted textually with every commit in this stack, but three of them touch the same function and would have merged to something that compiles while silently reordering locks or dropping the root-derivation ladder.

Reconciled by hand, not by taking either side wholesale:

- The empty-event early return now happens *after* `lock_settings_reload()` is taken, not before, so the branch never acquires `connections` (inside `apply_workspace_folder_change`) ahead of `reload` on any path — preserving main's invariant. The pool's own emptiness check still returns before it reaches `connections`, so nothing is held across it for a genuinely empty event.
- Root derivation uses `config_root_after_folder_change` (main's ladder) rather than this branch's original bare first-folder read, so the `--config-file`-absent path still gets the #746/#954 fallback behavior on a folder change.
- The `--config-file` branch keeps both effects main and this branch each added there — `set_root_path` (main) and the forced repull (this branch, since that branch returns before any reload and is its only chance) — a plain conflict-marker resolution would have kept only one.
- The commit that skipped the reload for an empty event cited a config-layer-drop defect that #746/#954 has since fixed independently; its message was reworded to keep the still-valid cost argument (avoiding a full reparse + semantic-tokens refresh for a no-op notification) without the now-stale defect citation.

Verified: `resolve_acquire`'s `Shared`-key invariant this branch's exemption relies on ("a `Shared` key is only ever minted for a marker-rooted acquisition") still holds on current main — the palette-command work landed in the interim reworks connection identity but not that gate. `apply_workspace_folder_change` in `pool.rs` was untouched by main, so its half of the rebase was mechanical.

## Validation

- `cargo test --lib` — 3362 passed, 0 failed (post-rebase)
- `cargo clippy --all-features -- -D warnings` — clean
- `cargo fmt --check` — clean
- Full e2e suite (`scripts/test_e2e_parallel.sh`) — 59/59 binaries, 1800 tests passed, 0 failing, including `e2e_downstream_refresh_skips_pull_fallback_disabled_contexts` and `e2e_workspace_folder_change_refreshes_pull_clients`.

An earlier revision of this description attributed a different e2e flake to this branch on the strength of a base-vs-branch comparison. That comparison was confounded by ambient load drifting between the two measurements and does not support the claim; it is withdrawn.

Every fix commit was verified RED on its parent first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Diagnostics now refresh automatically when workspace folders are added or removed.
  - Diagnostic results are invalidated after effective folder changes, preventing stale information.
  - Empty or unchanged folder notifications no longer trigger unnecessary reloads or refreshes.
  - Workspace connections remain stable when folder changes do not affect their configured scope.
  - Configuration-file-based workspace settings now resolve relative paths against the updated folder.

- **Tests**
  - Added coverage for diagnostic refreshes, connection stability, and configuration behavior during workspace-folder changes.


atusy:
## Follow-ups filed, both out of scope here

**#932** — `announce_shared_root` has this PR's defect for a different trigger: a marker root joining a `preferSharedInstance` connection changes its project and moves no generation either. It sits on the per-request acquisition path, where an unconditional bump would be a hot-path regression, and the fix needs a lock-ordering decision (`add_and_announce` commits under the folder-set lock).

**#937** — a fenced response still *reaches* the editor: `resolve_diagnostic_pull_report` refuses to store the baseline but returns a stale `Full` verbatim and a stale `Unchanged` as empty, and the upstream handler has no per-document pull supersession. So the repair is not ordered against the harmful response. Raised by codex review of this branch.

Three things kept it out of this PR:

1. It applies to all three fences, and the settings path — which has no repair at all — is the worst of them. Fixing it here would fix the least-affected one.
2. Ordering is not arbitrary in the common case: the stale pull is enqueued on the downstream FIFO before the fence-triggering notification, so the downstream answers it first and, for a single-region document, the editor's last write is the fresh one. Corruption needs a response-order inversion across regions or servers.
3. The obvious fix — suppress or supersede the stale response — is the shape this project already evaluated and rejected as a `content_epoch` version gate, on the grounds that dropping a stale response can leave the editor holding an indefinitely stale set with nothing to correct it. The preferred design (surface "stale lineage delivered" and force a trailing refresh after it) needs a cross-layer return-type change through 8+ call sites.

codex agreed on all three and downgraded it to a follow-up.

Net for this branch: the `Unchanged → empty` outcome is newly reachable here for a pull crossing the fence, where it previously contributed the stale-but-populated baseline. That is the cost of fencing at all, and it is why the forced refresh ships in the same PR rather than after it — without the refresh the change would be a straight regression for pull-namespace editors.

atusy:
## Test change from the codex round, and a flake measurement

**`8c1994e4c`** replaces the empty-event e2e with an in-process test. codex pointed out the e2e could not prove what it claimed: tower-lsp runs ingress handlers concurrently (`concurrency_level(64)`), so the `shutdown` request I used as a FIFO barrier can be handled *before* the notification ahead of it — and `shutdown` cancels the token that suppresses an already-admitted refresh. A regression that refreshed on empty events would still produce an empty watch list and pass. That is this repo's own documented tower-lsp concurrency rule, and I walked straight into it.

`refreshes_requested` is recorded synchronously ahead of every admission gate, so asserting on it in process cannot be hidden that way. The test then makes the same request directly and expects the counter to move, so a zero cannot come from a fixture whose capability gate was shut. Mutation-verified: removing the guard fails it in 5s. The positive wire path keeps its e2e.

## Flake measurement — reporting, not claiming innocence

`e2e_code_action::a_completed_reopen_releases_the_command_it_was_holding` failed on two of my full-suite runs. Matched sampling, both in fresh isolated target dirs:

| | isolation | full parallel suite |
|---|---|---|
| base `e75add289` | 10/10 pass | 6/6 pass |
| this branch | 10/10 pass | 6 pass / 2 fail |

So it is load-dependent on both, and I cannot show the base failing. What I can say about mechanism: nothing in this diff executes in that test's path — it exercises `pending_reopen` and the executeCommand barrier, with no workspace-folder event anywhere — and the test hinges on a re-open settling inside the 2s `REOPEN_WAIT` before the command is sent, which is exactly the shape that loses under contention. The plausible coupling is indirect: this branch lengthens the `e2e_push_diagnostics` binary, which reshuffles the parallel pool's schedule and changes which binaries contend with `e2e_code_action`.

I am flagging it rather than asserting it is pre-existing, because 0/6 versus 2/8 does not separate the two hypotheses. If a maintainer would rather I chase it before merge, say so — the next step would be pinning whether the failure is the barrier expiring or the wire ordering inverting, which the test distinguishes in its own assertions.

(Also observed earlier and separately: `e2e_downstream_refresh_skips_pull_fallback_disabled_contexts` flakes on a 600ms positive deadline, measured failing on this branch's base too. Untouched — that deadline is a real latency assertion, since the competing path in the mock sleeps 1000ms, so widening it would weaken what it tests.)

atusy:
## Correcting my earlier flake report

The comparison I posted above does not hold, and the mechanism I offered for it was speculation. Withdrawing both.

I measured the base at one point in the session and the branch at another, and the machine's **ambient** load drifted between them — rust-analyzer, Activity Monitor, a browser, none of it test-related. At a load average of ~101 on 10 cores I now see `a_completed_reopen_releases_the_command_it_was_holding` fail **2 in 4 runs in isolation**, with nothing of mine running. So "base 0/6 vs branch 2/8" separates nothing, and the suggestion that this branch lengthening `e2e_push_diagnostics` reshuffled the parallel schedule was a hypothesis I never tested.

What survives: the test has two log-polling waits, and the first one was worth fixing on its own merits.

**`0c795f8a8`** converts it. `await_initial_open` waited for the predecessor's *eager* open — a background fan-out nothing drives — to appear in the mock's wire log within 15s. It is now driven by a codeAction: the request path runs `ensure_document_opened` before queueing, on the same connection FIFO, so a response carrying actions is proof the didOpen arrived. That is the idiom every other caller in this file already uses. What the poll was really establishing (exactly one incarnation, a didOpen for that host on the wire) is now asserted once rather than raced. The mock is gated not to crash on the second host, since that host's codeAction now exists to open it rather than to kill the predecessor; the first host still triggers the crash and the "must have crashed once" assertion still holds.

The **second** wait cannot be converted the same way — the re-open sweep is serviced asynchronously and no client request drives it, so no response proves it finished. Filed as **#938** with the reasoning, including why the existing `KAKEHASHI_E2E_STALL_REOPEN_MS` precedent does *not* justify adding a completion hook here.

Note that this commit is test-hygiene unrelated to this PR's subject; it is here because it was blocking a clean read of the suite while reviewing.

Gate after it: `cargo test --lib` 3236 passed / 0 failed, clippy and fmt clean.

atusy:
@greptileai

atusy:
@coderabbitai review

atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
Closing per discussion (see conversation this PR came out of).

The core mechanism here — pull-diagnostic-lineage fencing + forced refresh on workspace-folder changes — rests on a theoretical risk: `previousResultId` is typically scoped to document version rather than project/settings identity in many downstream LSP server implementations, so a stale lineage could in principle survive a folder change undetected. That risk has never been observed with any downstream server this bridges to in practice, and neither this PR nor its follow-ups (#932, #937, #965) trace back to an actual bug report — all were derived analytically during review. Given the added complexity (a new lock-ordering invariant threaded through `did_change_workspace_folders_impl`, ~400 lines of production code across `pool.rs` and the handler, and four follow-up issues), closing this as YAGNI rather than merging it.

Two genuinely independent, validated pieces are being extracted into their own PRs rather than lost:

- **Empty-event skip** — `apply_workspace_folder_change` should short-circuit for a `workspace/didChangeWorkspaceFolders` event that names neither an addition nor a removal, instead of running a full settings reload + document reparse + semantic-tokens refresh for a notification that changed nothing. This is a real, unrelated cost fix with no dependency on the lineage-fence work.
- **Test hygiene** — the `tests/e2e_code_action.rs` 15s wire-log poll (now request-driven) and the matching `tests/bin/mock_formatter.rs` crash-gate fix. Entirely unrelated to diagnostics; it was only bundled here because it was blocking a clean review read.

Links to both follow shortly. Closing follow-up issues #932, #937, #965, #966 alongside this — all four are part of the same fence family and don't stand on their own without it. #945 (independent notification-ordering bug predating this PR) and #969 (independent settings-diff-before-reparse optimization) stay open.

PR #947, opened 2026-07-31
feat(lsp): scope custom methods under textDocument, with deprecated aliases

atusy:
## What

Moves the 41 document-scoped custom LSP methods under a `textDocument/` scope segment, so the vendor namespace mirrors LSP's own scope-first axis — the one that lets `textDocument/diagnostic` and `workspace/diagnostic` coexist.

```
kakehashi/node                    → kakehashi/textDocument/node
kakehashi/node/namedChildren      → kakehashi/textDocument/node/namedChildren
kakehashi/captures/full/delta     → kakehashi/textDocument/captures/full/delta
kakehashi/internal/effectiveConfiguration   (unchanged)
```

`effectiveConfiguration` keeps its name: its params are empty, so it is server-scoped, and `internal/` marks visibility rather than scope. A future server-scoped method has an obvious home at `kakehashi/workspace/…`, which is the point — under the flat scheme the two sides of a feature would have been siblings with nothing in the name distinguishing them.

**No client breaks.** Old names stay callable through a rewriting middleware that warns once per distinct name, to both the server log and the editor's LSP log.

Rationale and the two rejected alternatives (stay flat; drop the vendor prefix and extend the spec namespace as clangd does) are in `docs/architecture-decisions/custom-method-namespace.md`.

## Three things that are load-bearing

**The alias layer must wrap the ordering gate.** `IngressOrderGate` classifies by method name to assign per-document wire-order tickets and knows only the canonical spellings; below it, a deprecated call would arrive unrecognized, pass ungated, and read a tree missing an edit that preceded it on the wire. Both orders compile and both answer every request, so the mistake is invisible at the call site — the composition therefore lives in one function, `lsp::ingress_stack`, which `main.rs` calls and a unit test drives.

**The mapping is a frozen allowlist, not a prefix rule.** A blanket `kakehashi/` rewrite would corrupt `internal/effectiveConfiguration` and would invent deprecated spellings for methods that never had one. The list must not grow with new methods — but every entry must stay registered, so an e2e test walks all 41 old spellings asserting none has become `MethodNotFound`, with a never-registered name asserted to *be* `MethodNotFound` so the sweep cannot pass vacuously.

**The notice goes to `window/logMessage`.** `log::warn!` alone is invisible in a default run: env_logger takes its level from `RUST_LOG` and filters at `Error` when unset, and writes to the server's stderr rather than the editor's output — so the client authors who need to migrate would never have seen it. `showMessage` is wrong in the other direction (up to 41 popups).

## Testing

- `cargo test --lib` — 3254 passed, 0 failed
- `cargo test --features e2e --test 'test_*' --test 'e2e_*'` — 1749 passed, 0 failed
- `cargo clippy --all-features --lib --bins -- -D warnings`, `cargo fmt --check` — clean

Note that CI's bare `cargo test` skips every `e2e`-gated file, so the deprecation coverage above is a local gate only.

## Review

Ran `/revise`: ten parallel local reviewers, then codex twice (continued thread, then a fresh session) — both returned no comments. Findings fixed along the way included a stale-comment class every reviewer caught independently, three code comments stating mechanisms that do not exist, a test whose docstring claimed to pin the wiring while asserting against a stack it built itself, and a vacuous assertion that passed for the empty string. Writing the test for the client-visible notice also caught a harness trap: `send_request` discards notifications while awaiting the response, so a naive test would have failed against working code.

## Summary by CodeRabbit

* **New Features**
  * Custom document methods now use the consistent `textDocument` namespace for node and capture operations.
  * Legacy method names remain supported through automatic compatibility aliases.
  * Deprecated aliases display a one-time replacement warning in server and client logs.
* **Documentation**
  * Updated protocol, architecture, and language-feature documentation with the new method names, scoping guidance, and migration details.
* **Bug Fixes**
  * Improved request routing and ordering for renamed document-scoped methods while preserving request data and node references.


atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
Both fixed in 317a81ad5.

- **`node/range` ambiguity** — real, and the rename sharpened it: the Alternative-C paragraph used the name for a *rejected* bulk endpoint while the table two screens up defines it as the shipped per-node accessor. Reworded to say which one shipped.
- **Untagged fence** — tagged `text`; every other fence in the file already carries a language.

@coderabbitai review

atusy:
will revisit when I decide really need this

PR #950, opened 2026-08-06
feat(bridge): retire palette commands no configured server can serve

atusy:
Stacked on #827 — review that first; this PR's diff is only the three commits above it.

## Summary

A raw palette command name is registered with the editor on first advertisement and
was never taken back. A server removed from `languageServers` therefore left its
commands in the palette for the rest of the session: permanently inert, and after
#827 permanently silent about why.

This makes the registration reversible, and closes the gap #827's PR comment records.

## The hard part is WHEN, not how

Every connection-removal site in the pool — the workspace-folder change, the settings
reload, and the acquire-time invalidation — calls `pending_reopen.arm(&key)`. A purge
means *"this connection is coming back"*, not *"this name is gone"*. Retiring there
would churn the editor's palette on every settings reload and every respawn.

A settings change is the only event that can actually retire a name, and only when NO
server that ever advertised it resolves any more. That is the same notion of "gone"
the invalidation loop beside it already uses, so the two cannot drift into disagreeing
about whether a server still exists.

## Three pieces

**The registry learns which SERVERS advertised each name.** Server names rather than
connection keys, because the question is whether any currently-configured server could
still produce the command — which survives respawns, re-roots, and the loss of every
connection. (#827 narrowed the recorded *keys* to the client-fallback ones the
reconnect branch can revive, which deliberately does not answer this.)

**Registration identity is derived from the command name**, not allocated and stored,
so nothing has to be remembered between registering a name and retiring it. That costs
one `Registration` per name instead of one per batch — they still travel in a single
request — and buys individual retirement: `client/unregisterCapability` names a
registration, so a shared id would mean unregistering the batch and re-registering the
remainder.

**Whether the editor holds a registration is tracked separately from whether the entry
exists.** The entry outlives the registration on purpose. A retired name must still be
recognised by the decode gate — a client can fire one from a keybind or a stale cache,
and a name shaped like a routing token would otherwise be decoded and misrouted — and
it must be *re-registered* rather than skipped when a server advertises it again.

## Also closes the gap recorded on #827

A fresh codex review of #827 found that a recorded reconnect candidate is a claim
never checked against reality: servers A and B both advertise `source.fixAll`, B comes
back after an upgrade without it, A crashes, and the name is refused as ambiguous
although only A was ever a real candidate. The stale entry cannot be *reached* — the
exact-name check rejects it — but it still **counts**.

A handshake is the moment that claim can be checked, so `register` now reconciles: a
name the connection no longer advertises drops that key as a candidate. Only the
candidate list is reconciled; the server stays recorded and the name stays registered,
because another server may still advertise it and retirement is a config decision, not
one connection's capabilities.

## What codex review changed

A fresh codex pass found three cases where a name could die **without the config
changing at all** — the sweep above only asks whether a SERVER still resolves.

- **A server that came back with no `executeCommandProvider` was skipped entirely.**
  The recording site only ran when the handshake produced a command list, so the shape
  an upgrade takes when it drops the feature — no list at all — left the previous
  incarnation's claims standing. It now always reconciles, empty list included,
  because "no list" is news rather than the absence of news.
- **A configured server that merely stopped offering one command kept it forever.**
  After an in-place upgrade the server still resolves; only the command is gone. So
  reconciliation drops the SERVER from the name as well as the connection key — every
  instance of one server runs the same binary and advertises the same list — and a
  name that loses its last advertiser that way is retired on the spot.
- **`registered_upstream` was set even when nothing was sent.** A client that cannot
  accept dynamic registration, or a forwarding loop that has gone away, means the
  editor holds nothing; believing otherwise would later try to retire an id it never
  had. The flag now means "the editor was asked to hold this".

Deliberately not changed, recorded so it is not re-derived: the flag flips on ENQUEUE,
not on the editor's acknowledgement. Every user of this channel is fire-and-forget by
design, and both failure directions self-heal — a rejected registration is re-offered
the next time a server advertises the name, and a failed retirement is overwritten by
the next registration of the same derived id.

Codex also confirmed the derived ids cannot collide (concatenating the exact command
name is injective, and the prefix has no other use in the repo).

## What bot review changed

- **Retirement was not gated on the client capability** — registration is skipped when
  the editor did not advertise `workspace.executeCommand.dynamicRegistration`, so
  unregistering named ids it was never given. Both sides now ask through one accessor.
- **"Gone" was resolvability, not spawnability** — a server left in `languageServers`
  with `enabled = false` or an empty `cmd` will never advertise again, so its commands
  are as dead as a deleted server's; `resolve(server).is_some()` kept them registered
  and inert for the session.
- **Register and retire of the same derived id could race** — the channel delivers in
  order but each message is dispatched into its own task. One lock across the round
  trip, taken only by these two message kinds.

Refuted with the rationale already in the tree: "no retry on unregister failure". The
flag flips on enqueue rather than on the editor's acknowledgement because every user of
this channel is fire-and-forget, and both directions self-heal.

Two existing tests were passing for the wrong reason and are fixed: they drove
`propagate_settings` with `BridgeServerConfig::default()`, whose empty `cmd` makes it
UNspawnable, and with a pool whose editor never accepted dynamic registration at all.

## Validation

- `cargo test --lib` — **3285 passed, 0 failed**
- e2e: `e2e_code_action` 45/0, `e2e_host_bridge` 44/0, `e2e_organize_imports` 27/0.
  One suite run of four reported `concatenated_default_merges_actions_from_both_servers`
  failing; it passes alone 3/3 and the full suite passes 2/2 on re-run, and the test is
  a code-action merge unrelated to this diff. Recorded rather than dropped — this repo's
  e2e is known to fail under parallel load with a different set each time.
- `cargo clippy --lib -- -D warnings` clean; `--lib --tests` shows only the 12 errors
  pre-existing on `origin/main`
- `cargo fmt --check` clean

Mutation-verified — each of these passed the suite before its test existed:

| mutation | caught by |
|---|---|
| handshake reconciliation removed | `a_respawn_that_dropped_a_command_stops_being_a_candidate_for_it` |
| retirement never fires | 3 tests |
| retirement fires when *some* advertiser survives, not none | `a_name_one_surviving_server_still_advertises_is_not_retired` |
| a retired name is not re-registered when re-advertised | 6 tests |

Plus a pool-level pair: removing the last advertiser emits `UnregisterCommands`, and
an ordinary reload emits nothing — the churn this design exists to avoid.

## Summary by CodeRabbit

* **New Features**
  * Palette commands are automatically registered and retired as language server capabilities and settings change.
  * Commands are removed when servers disconnect, are disabled, deleted, or stop advertising them.
  * Registration remains consistent across reconnects and configuration updates.

* **Bug Fixes**
  * Prevented orphaned or duplicate editor commands.
  * Improved reliability when registering or removing multiple commands.

### DIFF
--- a/src/lsp/bridge/actor/reader.rs
+++ b/src/lsp/bridge/actor/reader.rs
@@ -202,6 +202,13 @@ pub(crate) enum UpstreamRequest {
     /// and fires them by raw name (#628 palette-fired commands). Fire-and-forget:
     /// no reply/cancel (the editor's ack is ignored; routing fails soft anyway).
     RegisterCommands { commands: Vec<String> },
+    /// Retire palette command names the editor should stop offering, via
+    /// `client/unregisterCapability` (#823 follow-up). Sent only when a settings
+    /// change leaves a name with no configured server that could advertise it —
+    /// NOT when a connection dies, since every removal site expects the same key
+    /// back and retiring there would churn the palette on each respawn.
+    /// Fire-and-forget for the same reason as its sibling.
+    UnregisterCommands { commands: Vec<String> },
     /// Bring `key`'s virtual documents up to date: its previous connection was
     /// purged and has now been replaced by a `Ready` process
     /// (respawn-reopen-derives-its-targets).
--- a/src/lsp/bridge/pool.rs
+++ b/src/lsp/bridge/pool.rs
@@ -545,6 +545,18 @@ impl LanguageServerPool {
         &self.command_origins
     }
 
+    /// Whether the editor will accept a dynamic `workspace/executeCommand`
+    /// registration — it does only if it advertised `dynamicRegistration` (LSP
+    /// spec). Both registering and RETIRING a palette command must ask, or a
+    /// retirement names ids the editor was never given.
+    fn client_supports_dynamic_command_registration(&self) -> bool {
+        self.client_capabilities()
+            .and_then(|caps| caps.workspace)
+            .and_then(|ws| ws.execute_command)
+            .and_then(|ec| ec.dynamic_registration)
+            .unwrap_or(false)
+    }
+
     /// Tell the USER, not just the log, that a request they invoked could not be
     /// served — as a `window/logMessage` at WARNING.
     ///
@@ -924,6 +936,37 @@ impl LanguageServerPool {
         drop(connections);
         for (key, handle) in stale_handles {
             shutdown_invalidated_connection(key, handle);
+        }
+        // A settings change is the ONLY event that can retire a palette command
+        // name. Connection removal cannot: every removal site above arms a
+        // re-open for the same key, so the connection is expected back, and
+        // retiring there would churn the editor's palette on each respawn.
+        //
+        // "Gone" is SPAWNABILITY, not mere presence in the config. A server left
+        // in `languageServers` with `enabled = false` or an empty `cmd` will
+        // never advertise anything again, so its commands are exactly as dead as
+        // a deleted server's — and asking only whether the entry resolves would
+        // leave them registered and inert for the rest of the session.
+        //
+        // Deliberately a different question from the invalidation loop above,
+        // which asks whether a LIVE process still matches its config. This one
+        // asks whether a process could ever exist.
+        let retired = self.command_origins.retire_unadvertisable(|server| {
+            resolve(server).is_some_and(|config| config.is_spawnable_with_wildcard(None))
+        });
+        // Gated exactly like registration: a client that cannot accept a dynamic
+        // registration was never sent one, so unregistering would name ids it
+        // never held.
+        if !retired.is_empty()
+            && self.client_supports_dynamic_command_registration()
+            && let Err(e) = self
+                .upstream_request_tx
+                .send(UpstreamRequest::UnregisterCommands { commands: retired })
+        {
+            log::warn!(
+                target: "kakehashi::bridge",
+                "Failed to queue palette-command retirement (forwarding loop gone): {e}"
+            );
         }
         pushed
     }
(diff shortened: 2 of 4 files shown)


atusy:
@coderabbitai review

atusy:
### Bot round 1 — addressed in `a8ef446a4`

Three real, one refuted.

**Fixed — retirement was not gated on the client capability** (codex P2, Copilot, qodo). Registration is skipped when the editor did not advertise `workspace.executeCommand.dynamicRegistration`, so unregistering named ids it was never given. Both sides now ask through one accessor so they cannot drift.

**Fixed — "gone" was resolvability, not spawnability** (CodeRabbit Major, qodo). A server left in `languageServers` with `enabled = false` or an empty `cmd` will never advertise anything again, so its commands are as dead as a deleted server's — but `resolve(server).is_some()` is true for it, and they stayed registered and inert for the session. Deliberately a *different* question from the invalidation loop beside it: that one asks whether a LIVE process still matches its config; this asks whether a process could ever exist. Coverage added for disabling the last advertiser.

**Fixed — register/retire of the same derived id could race** (codex P2). The channel delivers in order but each message is dispatched into its own task, so nothing downstream preserves it. One lock held across the round trip, taken only by these two message kinds, at handshake and reload rate.

**Also fixed** (Copilot): reconciliation built its "did this handshake advertise it" test as a linear scan per known name, under the lock, against a map that keeps every name for the session. One set instead.

**Refuted — "no retry on unregister failure"** (qodo). Deliberate, and already stated in the tree. The flag flips on enqueue rather than on the editor's acknowledgement because every user of this channel is fire-and-forget, and both directions self-heal: a rejected registration is re-offered the next time a server advertises the name, and a failed retirement is overwritten by the next registration of the same derived id.

**Partly refuted** (Copilot: skip reconciliation for non-fallback keys). Reconciliation must run for every key — it is what lets a server that stopped advertising a command retire it, which the config sweep cannot see. The perf half of that comment is the set change above.

Two existing tests were passing for the wrong reason and are fixed: they drove `propagate_settings` with `BridgeServerConfig::default()`, whose empty `cmd` makes it UNspawnable, and with a pool whose editor never accepted dynamic registration at all.

Gate: lib 3285/0, clippy `--lib -D warnings` clean, fmt clean.

atusy:
@coderabbitai review

atusy:
### Bot round 2 — addressed in `af7ae9d70`

One nitpick, and it was a real asymmetry worth closing: the two other sends on that path ask whether the editor accepts dynamic registrations and this one did not. It is currently redundant — with no capability nothing was announced, so nothing can be orphaned — but that invariant is two steps away, and a retirement that stops asking would name ids the editor never held.

Gate: lib 3285/0, clippy `--lib -D warnings` clean, fmt clean.

PR #951, opened 2026-08-06
feat(bridge): give every connection its own palette entry

atusy:
Stacked on #950 (which is stacked on #827) — this PR's diff is only the two commits above it.

## Summary

#827 made an ambiguous raw palette name **refuse** rather than guess. That is the only
answer that never runs a command in the wrong workspace, but it leaves the user with no
way to say which workspace they meant — and per-root pooling is on by default, so two
git repositories are enough to lose a server's palette commands for the session.

This is the escape hatch: **every connection gets its own palette entry.**

N connections advertising `ruff.fix` now produce one raw entry plus N routed entries:

```
ruff.fix                                    ← the readable one; refuses when ambiguous
kakehashi|m|ruff|file:///repo-a|ruff.fix    ← unambiguous
kakehashi|m|ruff|file:///repo-b|ruff.fix    ← unambiguous
```

The raw name is all the downstream gives us and it names no workspace. But kakehashi
already mints names that *do* name a connection — for code actions — so it mints them
here too. Picking a routed entry **is** the disambiguation: the choice moves from the
router, which has no information, to the user, who does.

## Both, not either

Registering only routed names would break every user's muscle memory and every keybind
holding a raw id, and would swap a readable palette entry for a machine-generated one
in the ordinary single-connection case where nothing was ambiguous to begin with.
Registering only raw names is where #827 left us. So: both. The readable entry keeps
working when it can, and the routed one is there when it cannot.

## The gate, now measured

`docs/architecture-decisions/execute-command-routing-token.md` deferred exactly this,
pending two questions **only a real client can answer**:

1. Do clients tolerate `workspace/executeCommand` registrations accumulating as roots
   are discovered lazily? (Per-connection naming makes them disjoint rather than
   overlapping, which shrinks the question but does not answer it.)
2. Does a client that filters on registered ids also expect a human-readable id? The
   protocol has no title field — `ExecuteCommandRegistrationOptions` carries only
   `commands` — so the palette label *is* the id.

**Measured in Neovim** (v0.13 nightly), with one kakehashi spanning two `.git` roots
so one downstream server has two connections. `scripts/probe_palette_registration.lua`
is checked in so the claim is re-runnable rather than described:

```
== every executeCommand registration, in arrival order ==
  [req 1] id=kakehashi/executeCommand/mock.run
  [req 1] id=kakehashi/executeCommand/kakehashi|m|mock-codeaction|file://<root>/repo-a|mock.run
  [req 2] id=kakehashi/executeCommand/kakehashi|m|mock-codeaction|file://<root>/repo-b|mock.run
duplicate registration ids across the session: none
handler errors: none

commands the client now holds (3):
  kakehashi|m|mock-codeaction|file://<root>/repo-a|mock.run
  kakehashi|m|mock-codeaction|file://<root>/repo-b|mock.run
  mock.run

RAW  mock.run   -> result=nil, editor told: [kakehashi] command "mock.run" was not run:
                   several live connections [mock-codeaction@…/repo-b,
                   mock-codeaction@…/repo-a] advertise it …
ROUTED repo-b   -> result={ executed = "mock.run" }
```

- **Q1 answered: yes.** Registrations accumulate one per root as it is discovered and
  Neovim accepts every one without error. The raw name is announced exactly once; each
  new root adds only its own routed entry; no duplicate ids — which is precisely what
  the derived-id scheme (#950) buys.
- **Q2 answered: yes, for Neovim.** It keys dynamic registrations by *capability* name,
  so they land in `client.registrations.executeCommandProvider` — enumerable, and the
  machine-generated ids are usable as-is. Firing the raw name is refused *and reported*;
  firing repo-b's routed name runs the command on exactly that connection. That is the
  whole stack demonstrated end to end in a real client.
- **And the finding that reframes the rest:** Neovim does **not** advertise
  `workspace.executeCommand.dynamicRegistration` at all, so none of this happens there
  unless the user sets that capability. The palette registration feature — this PR's and
  #628's alike — is inert in a default Neovim.

vscode-languageclient is still unmeasured. Reverting is one condition in `pool.rs` if a
client rejects them.

## Step 5: the collision registering them creates

The palette-first gate preferred a raw-registry hit over decoding, because an encoded
name could not be a palette entry — so a hit could only mean the palette command.
Registering routed names breaks that premise. A downstream can now advertise a raw id
byte-identical to an entry kakehashi minted for a **different** connection, and the two
readings name different processes with nothing to separate them.

That collision is refused, visibly, like every other unresolvable name. The test that
pinned the old rule is replaced by one that pins the new one, plus a sibling proving an
ordinary raw name still takes the palette path — the guard has to fire only on the real
collision.

## A defect in #950 that this exposed

#950 dropped the whole SERVER from a name during reconciliation, reasoning that every
instance of one server runs the same binary. True of the binary, wrong of the
bookkeeping: one connection's handshake speaks for itself, not for its siblings, so the
first of two roots to drop a command retired an entry the second still backed. Only
per-connection tracking makes that visible, and it is fixed here — the server is
dropped once none of its connections advertise the name.

## What codex review changed

Three findings, all consequences of registering routed names that the first pass did
not follow through:

- **The collision guard fired on decodability, not on a collision.** A downstream may
  advertise a raw id shaped like a routed name; that is only ambiguous if a routed
  entry of the same text actually exists. With nothing on the other side there is one
  reading and it routes fine. The guard now asks whether kakehashi *holds* that routed
  entry. Worth stating plainly: **the test I wrote for this pinned the false positive**
  — it set up a name that merely decoded and asserted a refusal, so it would have kept
  the bug alive. Replaced by one that mints the routed entry first, plus a sibling
  proving the non-collision still routes.
- **Settings retirement never cleared routed rows**, so removing a server left its
  routed entries visible and broken — and if another server still advertised the raw
  name, nothing was returned at all and the dead row leaked silently.
- **The announcement batch was not deduplicated**, so a downstream advertising both
  `foo` and the literal text of `foo`'s routed name would put two Registrations with
  the same derived id in one request, which a client may reject wholesale.

## What bot review changed

Two findings had already been fixed before the bots reported (they reviewed an earlier
commit): the collision guard testing decodability rather than a real routed twin, and
settings retirement leaving routed rows behind. The two that survived:

- **Routed entries were not rolled back when the request was never sent.**
  `forget_registration` took back the raw name but not the routed one that travelled
  with it, so the registry believed the editor held a row it never received.
- **`encoded_registered`'s comment claimed more than it tracks** — "the editor
  currently holds" was true only when every send succeeded.

## Validation

- `cargo test --lib` — **3291 passed, 0 failed**
- e2e — `e2e_code_action` 45/0, `e2e_host_bridge` 44/0, `e2e_organize_imports` 27/0
- clippy `--lib -D warnings` clean; `--lib --tests` only the 12 pre-existing on `main`
- `cargo fmt --check` clean

**The e2e earned its keep.** `downstream_command_names_are_registered_upstream_for_the_palette`
failed on the first attempt: the forwarding loop spawns a task per upstream request, so
sending the raw and routed names as two registrations made which one the editor saw
first a race. Both names were correct, both were sent, and every unit test passed —
only the wire order was wrong, which is the one thing a unit test of the registry
cannot see. They now travel in one request.

## Summary by CodeRabbit

- **New Features**
  - Added dynamic routing for encoded palette commands, allowing commands from multiple workspace connections to reach the correct destination.
  - Supports per-connection command registration and cleanup as servers become available or are removed.

- **Bug Fixes**
  - Prevented ambiguous command collisions by rejecting conflicting raw command names.
  - Improved cleanup when command registration fails or dynamic registration is unsupported.

- **Documentation**
  - Documented dynamic registration behavior and client opt-in requirements.

### DIFF
--- a/docs/architecture-decisions/execute-command-routing-token.md
+++ b/docs/architecture-decisions/execute-command-routing-token.md
@@ -245,21 +245,39 @@ narrows by configuration first.
 
 ## Decision–Implementation Gap
 
-**Dynamic registration of encoded names is deferred.** Making the name set
-finite is the precondition, not the feature. Two questions must be settled
-empirically before registering them, and neither can be answered from the
-protocol text:
+**Dynamic registration of encoded names is no longer deferred (#823 stack).**
+Making the name set finite was the precondition. Two questions had to be settled
+empirically first, and neither could be answered from the protocol text:
 
 - Roots are discovered lazily as documents open, so the registerable set grows
-  during a session. The existing palette registration mints a fresh
-  registration id per batch and never unregisters; whether real clients
-  tolerate repeated `workspace/executeCommand` registrations with overlapping
-  command lists is untested.
-- Whether a client that filters on registered ids also expects a
-  human-readable command id, given these are machine-generated.
+  during a session. Whether real clients tolerate repeated
+  `workspace/executeCommand` registrations, with the ids and lists that produces,
+  was untested.
+- Whether a client that filters on registered ids also expects a human-readable
+  command id, given these are machine-generated.
 
-Until that is settled, `executeCommandProvider.commands` stays empty and the
-vscode-languageclient limitation stands.
+**Measured in Neovim** (v0.13 nightly, `scripts/probe_palette_registration.lua`,
+one kakehashi spanning two `.git` roots so one downstream server has two
+connections):
+
+- Registrations DO accumulate — one request per root as it is discovered — and
+  Neovim accepts every one without error. The raw name is announced exactly once;
+  each new root adds only its own encoded entry. No duplicate registration ids,
+  which the derived-id scheme (one id per name, `kakehashi/executeCommand/<name>`)
+  is what guarantees.
+- Neovim keys dynamic registrations by CAPABILITY name, so they land in
+  `client.registrations.executeCommandProvider` — enumerable, and the encoded
+  names are usable as-is: firing one runs the command on exactly its connection,
+  while firing the raw name is refused with a `window/logMessage` naming the
+  connections that collided.
+- **Neovim does not advertise `workspace.executeCommand.dynamicRegistration` at
+  all** (checked against `vim.lsp.protocol.make_client_capabilities()`), so none
+  of this happens there unless the user opts in by setting that capability. The
+  palette-registration feature is inert in a default Neovim.
+
+Answered for Neovim only. The vscode-languageclient limitation is unmeasured, so
+`executeCommandProvider.commands` still stays empty in the initialize result —
+this path is dynamic registration only.
 
 **A DEAD shared-instance connection cannot be re-rooted from the token alone.**
 A `preferSharedInstance` connection is keyed without a root, and announcing a
--- a/scripts/probe_palette_registration.lua
+++ b/scripts/probe_palette_registration.lua
@@ -1,0 +1,209 @@
+-- Manual probe for the palette-command registration gate in
+-- docs/architecture-decisions/execute-command-routing-token.md.
+--
+-- The two questions that ADR deferred can only be answered by a real client:
+--   Q1 does the editor tolerate a `client/registerCapability` for
+--      `workspace/executeCommand` arriving repeatedly, as workspace roots are
+--      discovered lazily through a session?
+--   Q2 is a machine-generated command id usable where the client surfaces it?
+--
+-- Run from the repo root, after `cargo build --features e2e`:
+--
+--   PROBE_ROOT=/path/with/repo-a-and-repo-b \
+--   KAKEHASHI_BIN=target/debug/kakehashi \
+--   MOCK_BIN=target/debug/mock-lsp-formatter \
+--   FORCE_EXEC_DYNREG=1 \
+--     nvim --headless --noplugin -u NONE -l scripts/probe_palette_registration.lua
+--
+-- PROBE_ROOT must hold `repo-a/` and `repo-b/`, each with a `.git` directory
+-- and a `doc.md` containing a lua fence. Two marker roots is the whole point:
+-- per-root pooling is on by default, so ONE kakehashi spawns TWO connections of
+-- one downstream server -- the shape that makes a raw palette name ambiguous
+-- (#823).
+--
+-- FORCE_EXEC_DYNREG exists because Neovim does NOT advertise
+-- `workspace.executeCommand.dynamicRegistration` on its own, so without it
+-- kakehashi registers nothing and the probe measures nothing. That is itself a
+-- finding, recorded in the ADR.
+
+local root = vim.fn.fnamemodify(vim.env.PROBE_ROOT or vim.uv.cwd(), ":p"):gsub("/$", "")
+local kakehashi = vim.env.KAKEHASHI_BIN
+local mock = vim.env.MOCK_BIN
+
+local log = {}
+local function say(fmt, ...)
+  local line = select("#", ...) > 0 and string.format(fmt, ...) or fmt
+  table.insert(log, line)
+end
+
+-- Capture every registerCapability / unregisterCapability the server sends,
+-- then hand off to Neovim's own handler so we also learn whether IT accepts them.
+local registrations = {}
+local unregistrations = {}
+local handler_errors = {}
+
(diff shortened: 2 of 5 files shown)


atusy:
@coderabbitai review

atusy:
@codex review

atusy:
### Bot round 1 — addressed in `c6afb1eb5` and `a5e302da9`

Two of the four had already been fixed in `c6afb1eb5` before the bots reported (they reviewed `b8090e0ca2`):

- **collision guard too broad** (codex P2, qodo) — refusing on decodability rather than on a routed twin that actually exists. The test I had written for it pinned the false positive; it is replaced by one that mints the routed entry first, plus a sibling proving the non-collision still routes.
- **routed rows not retired when a server is removed** (codex P2) — routed entries name a connection, so they now retire with their own server, independently of the raw name's fate.

The two that survived, fixed in `a5e302da9`:

- **Routed entries were not rolled back when the request was never sent** (qodo, Copilot). `forget_registration` took back the raw name but not the routed one that travelled with it, so the registry believed the editor held a row it never received and would never re-offer it.
- **`encoded_registered`'s comment claimed more than it tracks** (Copilot). It said "the editor currently holds", true only when every send succeeded. It now says what it is: the connections the editor was *asked* to hold.

### The ADR gate is now measured, in Neovim

`scripts/probe_palette_registration.lua` is checked in so this is re-runnable rather than described. One kakehashi spanning two `.git` roots, so one downstream server has two connections:

```
[req 1] id=kakehashi/executeCommand/mock.run
[req 1] id=kakehashi/executeCommand/kakehashi|m|mock-codeaction|file://<root>/repo-a|mock.run
[req 2] id=kakehashi/executeCommand/kakehashi|m|mock-codeaction|file://<root>/repo-b|mock.run
duplicate registration ids across the session: none
handler errors: none

RAW  mock.run -> result=nil, editor told:
  [kakehashi] command "mock.run" was not run: several live connections
  [mock-codeaction@…/repo-b, mock-codeaction@…/repo-a] advertise it …
ROUTED repo-b -> result={ executed = "mock.run" }
```

- **Q1 yes** — registrations accumulate one per root and Neovim accepts every one without error; the raw name is announced exactly once, each root adds only its own routed entry, no duplicate ids.
- **Q2 yes, for Neovim** — it keys dynamic registrations by *capability* name, so they land in `client.registrations.executeCommandProvider`, enumerable, and the machine-generated ids work as-is.
- **And the finding that reframes the rest:** Neovim does not advertise `workspace.executeCommand.dynamicRegistration` at all, so none of this happens there unless the user sets it. This feature — and #628's — is inert in a default Neovim.

vscode-languageclient remains unmeasured.

Gate: lib 3291/0, e2e 45/44/27, clippy `--lib -D warnings` clean, fmt clean.

atusy:
@coderabbitai review

atusy:
### Bot round 2 — addressed in `2e1943569`

Both Majors were real.

**Routed-only announcements were never rolled back.** When a second connection advertises a name the first already announced, it contributes a routed entry and NO raw one — so keying the rollback off the raw list passed an empty slice and silently kept a routed row the editor never received, which nothing would ever re-offer. The two halves are taken back separately now, because they are independent: a raw name may be shared, so a routed failure must not clear it, and vice versa.

**A raw command id can equal a routed name, and the registration id is derived from the final text** — so both owners shared one id and either retiring would unregister what the other still needs. Neither is announced twice now, and the check is against the **whole registry** rather than the current handshake, because the two owners need not come from the same connection or even the same handshake. (Your suggestion was reference counting by final name; withholding the second owner is the smaller move and reaches the same invariant.)

At most one owner is the property; *which* one depends on advertisement order and does not matter — the survivor is unusable either way, since dispatch refuses that exact text. Nothing the user could have run is lost. That also makes the wire-level dedup unreachable, so it and its precondition test are gone: the invariant now lives where the decision is made.

**Probe fix taken too**, and it mattered more than it looked: waiting only for the request callback filed the refusal notification under the *next* command, so the output reported the raw refusal as silent and the routed success as warned — the exact opposite of what happens. Re-measured in Neovim after the fix; the result is unchanged.

Gate: lib 3292/0, e2e 45/44/27, clippy `--lib -D warnings` clean, fmt clean.

PR #961, opened 2026-08-06
docs(adr): workspace-scoped symbol search

atusy:
Design record for `workspace/symbol` — the first bridged request that carries no `textDocument`, which breaks all three mechanisms every existing handler relies on.

## Why this needed its own decision

Every bridged request today resolves a host document → an injection region → a `RegionOffset`, and that chain is what selects servers (by injection language), filters results (against *the* request's virtual URI), and arbitrates between them (first-non-empty wins). A workspace query has none of it.

The subtler problem is arbitration: **a `workspace/symbol` response is not attributable to a language.** A server answers "here are the workspace's symbols matching `query`" and what it indexes is its root, not a language — so a server reached through one configured pair routinely returns another pair's symbols. Nothing in the response says which pair owns an entry, so no arbitration between pairs can be principled. That single fact drives most of what follows.

## What it decides

- **`union` everywhere, `preferred` overridden.** Two servers' symbol sets are complementary, not competing — the repo's own criterion, applied. Collect from every target, dedup on `(name, kind, uri, range)`, sort on a superset of that key.
- **Coverage is what is live.** A query never cold-starts a server; spawning buys *zero* embedded-block coverage, since virtual documents reach a downstream only via `didOpen` of host documents the client opened.
- **Cross-block, deliberately.** Each result resolves its own region's geometry, which is exactly what the goto filter lacks — and why goto must drop cross-region results.
- **Split by visibility**: the send lives in `bridge` (`connections()` is `pub(super)` there); fan-in lives in `lsp_impl` by coupling, not constraint.
- **Every entry ends in one of three states** — translated, semantically rejected, infrastructure-failed — and the outcome rolls up from those. The line: semantic rejection is *a place that no longer exists*; infrastructure failure is *output that cannot be used*. An all-failed query returns an error, never a confident `[]`.

## Two shipping blockers

§9 names two gaps as **blockers, not accepted limitations** — and deliberately does not specify their mechanisms:

1. A connection must outlive the documents that justified it. `didClose` leaves the connection running while removing document tracking, so closing the last buffer for a language drops a still-`Ready` server; the respawn window and an all-failed-startup workspace are the same defect from other angles.
2. This query's total fan-out must be bounded. `max_fan_out` selects server *names* per pair and does not cap the query.

Both **were** specified here (a pool-held admission registry; a new top-level ceiling) and that was rolled back. Across four review rounds the findings against those two mechanisms grew rather than shrank, and nearly all were runtime-lifecycle questions — invalidation distinguishing a failure respawn from a reconfiguration, snapshots outliving the invalidation that cleared them, cap slots spent on connections replaced before enqueue. Those are settled by tests, not prose. Considered Options records the attempt so it is not repeated by accident.

## Review

`/review` (10 perspectives, 4 rounds) then codex MCP across seven independent fresh sessions. A fresh session now returns clean on its first response.

Substantive corrections the review forced, among others: `preferred` never short-circuits sending; `resolve_virtual_uri` is an O(N) scan whose own doc comment justifies itself on rarity; `wait_for_response` has a hardcoded 30s timeout; the workspace-membership test must be **overlap**, not containment (`.git` above the opened folder is the ordinary case); and `resolve_by_region_id` *mutates* the tracker, so fan-in must never resolve inline.

Follow-up work is tracked as tasks, including the two blockers.

### DIFF
--- a/docs/architecture-decisions/workspace-scoped-symbol-search.md
+++ b/docs/architecture-decisions/workspace-scoped-symbol-search.md
@@ -1005,7 +1005,7 @@ normalization**: set `deprecated = Some(true)` when the normalized tags contain
 it was meant to preserve. It is bounded: the modern type is used everywhere
 else.
 
-### 6. Latency is bounded by the pool's timeouts and our own phase budgets
+### 6. Every wait is bounded; the synchronous work between them is not
 
 Every target is awaited, so latency is max-over-targets. The bounds are:
 
@@ -1105,8 +1105,8 @@ one region's offset is known. Workspace symbol search holds every region's
 geometry for the hosts it touches, so each entry is translated by its own
 region's offset rather than by a single borrowed one.
 
-Shipping this obliges edits in **both** user-facing docs, in three separate
-respects each. They are listed here because more than one review round found the
+Shipping this obliges edits in **both** user-facing docs, in four separate
+respects. They are listed here because more than one review round found the
 inventory incomplete.
 
 *The no-cross-block rule is stated as a blanket claim and must be amended, not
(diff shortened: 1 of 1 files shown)


atusy:
@coderabbitai review

atusy:
@greptileai

PR #962, opened 2026-08-06
feat(install): make installing a language all-or-nothing

atusy:
## Why

`kakehashi language install` installed the parser and the query files independently, so either half could land without the other:

- A failed queries download left a parser with no highlighting; a failed compile left queries with no parser.
- Re-running to repair the missing half printed `✗ Parser already exists … Use --force` for the half that *was* fine, and exited 1 — even though it had just repaired the other one. A plain re-run of a fully installed language could never succeed, which is why `Makefile` passes `--force` on every install.
- A `; inherits:` parent that failed to download was a warning, so a language could be published with a dangling inherits chain that fails to load at runtime.

## What

Both halves are **staged** first — the parser compiled to a temp file, the queries (and every language they reach through `; inherits:`) downloaded into per-language staging directories — and published only once all of them are ready. Queries are staged first because they are the cheap half: an unsupported language now fails in seconds instead of after a parser compile that gets thrown away. The parser is published last, since request routing gates on a registered parser.

A per-language lock orders a whole install against another install or an uninstall, taken for every language in the dependency chain in sorted order. Publishing a query directory keeps the one it displaced until the install commits, so a failure can put it back.

The CLI reports what the data directory holds rather than which half this run wrote, so an already-installed half is no longer an error.

## Behavior changes

- A `; inherits:` parent that cannot be downloaded now **fails the install** instead of warning. Maintainer decision: a language whose inherited queries are missing does not load.
- A language whose queries do not exist upstream can no longer get its parser installed. Measured against current nvim-treesitter: 0 of 323 parsers lack `highlights.scm`, so nothing is affected today.
- `kakehashi language install <already-installed>` now exits 0 instead of 1. Scripts that used exit 1 to detect "already installed" will see success.
- Auto-install no longer skips a language on the strength of its parser alone; a data directory left parser-only by an older release repairs itself on open.

## Known limitations (documented in `docs/README.md`)

- A `--force` reinstall killed between the two publishes leaves new queries against the old parser; both halves look installed, so recovery is another `--force`.
- Publication is atomic per file, not to readers: a server loading a language *while* it is being published can cache a failed query load, and recovery is install-then-reload.
- Base languages are tracked by name, not by query kind, so a base language whose `injections.scm` upstream does not ship (or whose optional download failed) still passes.
- Staging-file recovery is a no-op on Windows, where process liveness cannot be tested portably.

## Review

Ten parallel single-perspective reviews plus nine rounds against the codex MCP server (two independent sessions, the second cold). Every finding was fixed or is listed above; the fixes are separate commits naming the finding. One finding was caught by an existing CLI test — `test_language_status_preserves_incomplete_query_dir_when_backup_exists` — which pinned the invariant a new backup sweep would have broken.

## Verification

- `make check` clean
- `cargo test --lib` 3325 pass, `cargo test --bins` 7 pass
- `make test_e2e` 1764 pass / 0 failing binaries
- Manual: clean install, repair from either half missing, re-install of a complete language, unsupported language, invalid name, install→uninstall→install, and a language with an inherits chain (`typescript` → `ecma`)

## Summary by CodeRabbit

* **New Features**
  * Language installations now publish parsers and inherited queries transactionally.
  * Added recovery for interrupted or partially failed installations.
  * Coordinated concurrent installs and uninstalls to prevent conflicts.
  * Installation results report downloaded files, warnings, and rollback status.

* **Bug Fixes**
  * Prevented incomplete installations from being reported as successful.
  * Improved retries, temporary-file cleanup, and missing parser/query recovery.
  * Preserved parser files when query removal fails.
  * Improved handling of installation races and inherited queries.

* **Documentation**
  * Expanded guidance for retries, force reinstalls, cleanup, base-language queries, and troubleshooting.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -874,6 +874,14 @@ kakehashi language uninstall lua --force
 kakehashi language uninstall --all --force
 ```
 
+Installing a language is all-or-nothing. The parser, the language's query
+files, and every query file it pulls in through `; inherits:` are prepared
+first, and only made visible once all of them are ready — so a failed install
+(no network, an unsupported language, a compile error) leaves the data
+directory exactly as it was. Re-running the command is always safe and never
+needs `--force`: whatever is already installed is left alone, and only the
+missing pieces are fetched.
+
 ### Configuration Management
 
 ```bash
@@ -1173,9 +1181,10 @@ RUST_LOG=kakehashi::lock_recovery=warn kakehashi
 
 ### Queries not working for TypeScript/JavaScript
 
-These languages use query inheritance. Ensure base queries are installed:
+These languages use query inheritance. Re-run the install to fetch any base
+queries that are missing:
 
 ```bash
-kakehashi language install typescript --force
-# This automatically installs 'ecma' queries
+kakehashi language install typescript
+# This also installs the 'ecma' queries typescript inherits from
 ```
--- a/docs/README.md
+++ b/docs/README.md
@@ -897,7 +897,9 @@ languages are never replaced, so install one by name to refresh it.
 One gap `--force` is the only fix for: a `--force` reinstall that is killed
 between publishing the new queries and publishing the new parser leaves new
 queries against the old parser. Both halves look installed, so a plain re-run
-accepts them — reinstall with `--force` to line them up again.
+accepts them — reinstall with `--force` to line them up again. A kill that
+leaves the parser itself missing needs no `--force`: a plain re-run sees the
+missing half and fetches it.
 
 ### Configuration Management
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -901,6 +901,19 @@ accepts them — reinstall with `--force` to line them up again. A kill that
 leaves the parser itself missing needs no `--force`: a plain re-run sees the
 missing half and fetches it.
 
+Two more limits worth knowing:
+
+- Publication is atomic per file, not to readers. A running server that loads a
+  language *while* a `--force` reinstall is publishing can see the new queries
+  against the parser still being replaced, and keep that pairing until it
+  reloads. Installing while nothing is using the language avoids it.
+- A base language is fetched with the same query kinds as the language that
+  inherits it, but only the ones upstream publishes for it. If upstream names a
+  base language in, say, `injections.scm` without shipping that base language's
+  `injections.scm`, the install still succeeds and the query fails to load —
+  the missing file is upstream's, and installing the base language by name will
+  not conjure it either.
+
 ### Configuration Management
 
 ```bash
--- a/docs/README.md
+++ b/docs/README.md
@@ -876,11 +876,17 @@ kakehashi language uninstall --all --force
 
 Installing a language is all-or-nothing. The parser, the language's query
 files, and every query file it pulls in through `; inherits:` are prepared
-first, and only made visible once all of them are ready — so a failed install
-(no network, an unsupported language, a compile error) leaves the data
-directory exactly as it was. Re-running the command is always safe and never
-needs `--force`: whatever is already installed is left alone, and only the
-missing pieces are fetched.
+first, and only made visible once all of them are ready. A failed install — no
+network, an unsupported language, a compile error, or a base language that
+cannot be downloaded — publishes neither half, so whatever was installed before
+is still exactly what is installed. (Bookkeeping is not covered by that: the
+parser metadata cache, the `parser/` and `queries/` directories, and the
+per-language lock files may be created or refreshed either way.)
+
+Re-running the command is safe: whatever is already installed is left alone,
+and the missing halves — including the queries of a base language it inherits —
+are fetched without `--force`. Use `--force` to *replace* something already on
+disk, such as a parser that fails to load.
 
 ### Configuration Management
 
@@ -1182,7 +1188,8 @@ RUST_LOG=kakehashi::lock_recovery=warn kakehashi
 ### Queries not working for TypeScript/JavaScript
 
 These languages use query inheritance. Re-run the install to fetch any base
-queries that are missing:
+queries that are missing; if a base language cannot be downloaded the install
+fails rather than leaving a dangling `; inherits:`:
 
 ```bash
 kakehashi language install typescript
--- a/docs/README.md
+++ b/docs/README.md
@@ -878,15 +878,21 @@ Installing a language is all-or-nothing. The parser, the language's query
 files, and every query file it pulls in through `; inherits:` are prepared
 first, and only made visible once all of them are ready. A failed install — no
 network, an unsupported language, a compile error, or a base language that
-cannot be downloaded — publishes neither half, so whatever was installed before
-is still exactly what is installed. (Bookkeeping is not covered by that: the
-parser metadata cache, the `parser/` and `queries/` directories, and the
-per-language lock files may be created or refreshed either way.)
+cannot be downloaded — never leaves the language you asked for half-installed:
+its parser and its queries are published together or not at all.
+
+Two things a failed install can still leave behind. Queries for a *base*
+language it had to fetch stay installed: they are shared with every language
+that inherits them, and a base language nothing uses is inert. And bookkeeping —
+the parser metadata cache, the `parser/` and `queries/` directories, the
+per-language lock files, and the clearing of the language's uninstall
+tombstone — is written either way.
 
 Re-running the command is safe: whatever is already installed is left alone,
 and the missing halves — including the queries of a base language it inherits —
-are fetched without `--force`. Use `--force` to *replace* something already on
-disk, such as a parser that fails to load.
+are fetched without `--force`. Use `--force` to *replace* the requested
+language's parser and queries, such as a parser that fails to load; base
+languages are never replaced, so install one by name to refresh it.
 
 ### Configuration Management
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -878,8 +878,10 @@ Installing a language is all-or-nothing. The parser, the language's query
 files, and every query file it pulls in through `; inherits:` are prepared
 first, and only made visible once all of them are ready. A failed install — no
 network, an unsupported language, a compile error, or a base language that
-cannot be downloaded — never leaves the language you asked for half-installed:
-its parser and its queries are published together or not at all.
+cannot be downloaded — does not leave the language you asked for
+half-installed: its parser and its queries are published together or not at
+all. (If undoing a publish itself fails — a permission error, a file held open
+— the command says so instead, and names what is left where.)
 
 Two things a failed install can still leave behind. Queries for a *base*
 language it had to fetch stay installed: they are shared with every language
@@ -901,18 +903,28 @@ accepts them — reinstall with `--force` to line them up again. A kill that
 leaves the parser itself missing needs no `--force`: a plain re-run sees the
 missing half and fetches it.
 
-Two more limits worth knowing:
+Three more limits worth knowing, all of them about a language being *loaded*
+while it is being installed, or about files upstream does not provide:
 
 - Publication is atomic per file, not to readers. A running server that loads a
   language *while* a `--force` reinstall is publishing can see the new queries
   against the parser still being replaced, and keep that pairing until it
-  reloads. Installing while nothing is using the language avoids it.
-- A base language is fetched with the same query kinds as the language that
-  inherits it, but only the ones upstream publishes for it. If upstream names a
-  base language in, say, `injections.scm` without shipping that base language's
-  `injections.scm`, the install still succeeds and the query fails to load —
-  the missing file is upstream's, and installing the base language by name will
-  not conjure it either.
+  reloads. Worse, a server that loads a language whose base language is being
+  removed at that moment registers it with the query error as a warning and
+  caches that verdict, so later opens do not retry the install. Both are fixed
+  by restarting the server or installing the language explicitly; installing
+  while nothing is using the language avoids them.
+- Base languages are tracked by name, not by which query kind needed them. A
+  base language is fetched with whatever kinds upstream publishes for it, and
+  the optional kinds are best-effort: if upstream does not ship the base
+  language's `injections.scm`, or that one download fails, a language whose own
+  `injections.scm` inherits it installs successfully and fails to load its
+  injections. A plain re-run skips the base language as complete —
+  `kakehashi language install <base> --force` is what re-fetches it.
+- Staging files left behind by a process that was killed are collected by
+  `kakehashi language status` on Linux and macOS. On Windows there is no
+  portable way to tell whether the owning process is still running, so they are
+  left alone; they are inert, and safe to delete by hand.
 
 ### Configuration Management
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -894,6 +894,11 @@ are fetched without `--force`. Use `--force` to *replace* the requested
 language's parser and queries, such as a parser that fails to load; base
 languages are never replaced, so install one by name to refresh it.
 
+One gap `--force` is the only fix for: a `--force` reinstall that is killed
+between publishing the new queries and publishing the new parser leaves new
+queries against the old parser. Both halves look installed, so a plain re-run
+accepts them — reinstall with `--force` to line them up again.
+
 ### Configuration Management
 
 ```bash
--- a/docs/README.md
+++ b/docs/README.md
@@ -911,20 +911,25 @@ while it is being installed, or about files upstream does not provide:
   against the parser still being replaced, and keep that pairing until it
   reloads. Worse, a server that loads a language whose base language is being
   removed at that moment registers it with the query error as a warning and
-  caches that verdict, so later opens do not retry the install. Both are fixed
-  by restarting the server or installing the language explicitly; installing
-  while nothing is using the language avoids them.
+  caches that verdict, so later opens do not retry the install. Fixing that
+  takes both steps: install the missing language, then restart or reload the
+  server, since the install repairs the disk and the reload clears the cached
+  verdict. Installing while nothing is using the language avoids the whole
+  situation.
 - Base languages are tracked by name, not by which query kind needed them. A
   base language is fetched with whatever kinds upstream publishes for it, and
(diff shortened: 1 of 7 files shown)


atusy:
@coderabbitai review

atusy:
Fixed in `54e7d32ab`:

- **qodo** — `normalize_inherited_language_name` now strips only a *matched* pair of parentheses, character for character what `query_loader::normalize_inherited_language_name` does. Your read is right: on `(cpp` the loader looks for a language literally named `(cpp`, so fetching `cpp` would have reported success over something still unloadable. Tests pin `(cpp)` and both unmatched forms.
- **codex** — real regression, thank you. `try_lock_language` returns `LanguageLockProbe::{Idle, Busy, Unlockable}`; a data directory that cannot be written cannot have a publish in flight, so it reads as settled and preinstalled languages load again. Test covers a read-only `queries/`.
- **Copilot (parser.rs:307)** — the Windows displacement is gated on `is_file()` now. A directory sitting where the parser belongs is no longer moved aside and then stranded by a `remove_file` that cannot remove it; the rename fails and surfaces the error instead.
- **Copilot / CodeRabbit (queries.rs:808)** — the staging doc said "appends before recursing", which is the opposite of what it does since publication became dependency-ordered. Corrected.
- **Copilot (queries.rs:467)** — `&&entry.language == language` is now `entry.language == **language`.

Verification: `make check` clean, `cargo test --lib` 3327 pass, `test_cli` 63 pass, `make test_e2e` 1764 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
@codex review

atusy:
Round 2 fixed in `af224bf61`.

**Copilot, codex and qodo all found the same real defect, independently** — staging cleared an inherited base language's uninstall tombstone holding only the per-directory replace lock, so the clear could land *inside* a running `language uninstall <base>`, between the marker that uninstall writes and the parser it removes. The uninstall reported success and the install republished what it had just removed.

`clear_uninstall_tombstone_for_dependency` now takes that base language's own `LanguageLock` first, so the two are ordered: either the uninstall finishes and this install knowingly refetches the base language it needs, or it supersedes this install through the tombstone check the publish makes under the same lock. The requested language's clear has worked this way since the transaction lock landed; its dependencies now match. Test pins that the clear takes the lock and gives it back.

**Copilot (suppressed #2)** — also fixed: `InstallResult`'s doc said a `None` path with no error means that half is not published, which reads as a claim about the disk. A failure before staging leaves an earlier install's parser exactly where it was. The doc now says the paths describe what the run settled on, not what is on disk.

Verification: `make check` clean, `cargo test --lib` 3328 pass, `test_cli` 63 pass, `make test_e2e` 1764 pass / 0 failing binaries.

Note for anyone watching checks: `ci.yaml` has not fired for this PR — it has not fired anywhere in the repo since 15:01Z, including other open PRs, so it looks like an Actions-side condition rather than anything about this branch. The gate above was run locally, and it is stricter than CI (which runs no e2e).

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
@greptileai

atusy:
Round 3 fixed in `141455862`.

- **codex** — right, and it was the one publish failure that is not residue-free. `publish_query_dir` moves the live directory aside before renaming the staged one in; if both that rename and the restore fail, the language has no queries and the previous ones are in a backup, yet the install reported it like any other failure. A typed `QueryInstallError::PreviousQueriesStranded { language, backup, reason }` now carries the location, and the installer maps it to the `PreviousQueriesStranded` residue the CLI already reports. Reaching the branch needs two renames to fail in a row, so the test pins the part that matters to whoever hits it — the message names the backup.
- **qodo** — fixed. The Windows displacement now loops until it finds an unused `.displaced` name instead of trusting a generated one. Your reasoning about why it matters *there specifically* is the point: pids are reused and the staging sweep does not run off unix, so a stale file from a crashed process with the same pid would have failed every later reinstall at that rename.

Verification: `make check` clean, `cargo test --lib` 3329 pass, `test_cli` 63 pass, `make test_e2e` 1764 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 4 fixed in `2270b7315`.

- **codex** — you were right to raise this again; I had scoped it out earlier and the reasoning does not hold. `lock_complete_chain` now takes each language in the chain as it walks and holds them all until the caller is done, so the parser read cannot straddle a publish either. Still non-blocking (async path), and because nothing waits, discovery order cannot deadlock. It *replaces* the unlocked chain walk rather than sitting beside it — two walks of the same rule would have drifted apart.
- **Copilot** — fixed. `try_lock_language` returns a new `UnusableName` variant for a name this module could never have installed under, and `is_settled()` is now `Idle | Unlockable` rather than "not Busy", so an invalid name cannot be treated as settled state to read paths from.
- **qodo** — fixed. A dependency removed after staging reports `DependencyRemoved`, not `LanguageNotSupported`. Your remediation-path argument is the reason: the old error sent the user looking for a language nvim-treesitter does not have, when the remedy is simply to retry.

Verification: `make check` clean, `cargo test --lib` 3330 pass, `test_cli` 63 pass, `make test_e2e` 1764 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 5 fixed in `87e32c29c`. Copilot reported no new comments this round.

- **qodo** — real, and it defeated the residue reporting added two rounds ago. `StagedQueryInstall::publish` rolled back on its own error paths and discarded the `RollbackOutcome`. It now returns `PublishFailure { error, residue }`, and the installer folds that into `InstallResult::rollback_residue` alongside the `PreviousQueriesStranded` case. The two travel together because the caller reports them together.
- **codex** — fixed, and the failure mode you describe is the worse half: on a mount without advisory locks, `try_lock` errors rather than blocks, every language read as unusable, *and* the repair that triggered failed the same way. Only `TryLockError::WouldBlock` is `Busy` now; `TryLockError::Error(_)` is `Unlockable` — nobody can lock there, so nobody can publish there.
- **Copilot (suppressed)** — the `SuccessWithWarnings` log losing the parser/queries prefix is a fair readability point, but that branch is only reachable when another install completed the language while this one failed, and both error strings already name their own subject ("Parser installation failed…" / "…queries…"). Leaving it rather than adding a prefix that would often read as a double label.

Verification: `make check` clean, `cargo test --lib` 3330 pass, `test_cli` 63 pass, `make test_e2e` 1764 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 6 fixed in `dfe2bc7a7`. Copilot and codex reported nothing new this round.

- **qodo** — correct, and it was my own inconsistency: uninstall removed the parser whether or not the query removal had succeeded, so an I/O or permissions failure on the queries turned a failed uninstall into exactly the half-removed language this path exists to prevent. It now leaves both halves in place and says so, which keeps the retry a plain retry. New CLI test `test_language_uninstall_keeps_the_parser_when_queries_cannot_be_removed`.

Verification: `make check` clean, `cargo test --lib` 3330 pass, `--bins` 7 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 7 in `fd10ad681`. Two fixed, one refuted.

- **codex** — right, and it exposed that last round's rule was too blunt. The removal can take the query directory and *then* fail enumerating or deleting a backup; stopping there left a parser-only language, which is what the rule existed to prevent. `remove_query_install_and_backups` now returns `QueryRemovalFailure { error, removal }`, so uninstall finishes the job when the queries are already gone and stands down when they are not.
- **qodo (backup sweep hides io errors)** — fixed. `collect_superseded_backups` propagates per-entry read errors and uses `file_type()?` instead of `is_dir()`, matching what the parser sweep already did.
- **qodo (busy lock triggers wasted work)** — not changing this, and here is why. The waste is real but bounded to one staging+compile, and it needs two *processes* installing the same language at once: the in-process claim (`InstallingLanguages`) already dedupes the common case. The alternatives are worse. Treating `Busy` as "complete" would skip a repair the language may genuinely need; blocking on the lock would move a multi-minute wait onto the LSP's async path, which is the reason the probe is non-blocking at all. And the wasted work is self-limiting: the staged parser is dropped by `yield_to_existing` and the staged queries by `AlreadyComplete`, so the loser publishes nothing rather than fighting the winner. I would rather pay a rare duplicate compile than either miss a repair or stall the server.

Verification: `make check` clean, `cargo test --lib` 3330 pass, `--bins` 7 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 8 in `848332bfd`. Two fixed, one scoped.

- **codex** — right again, and it caught a distinction I had blurred: `removed_queries` says what the call *did*, not what it *left*. A language that was already parser-only reports no removal, so a later failure made uninstall keep the parser and change nothing at all. `QueryRemoval` now carries `queries_absent` — removed by this call or already gone — and the parser decision keys on that.
- **qodo (inherits read errors)** — fixed. `inherited_languages_on_disk` returns `Option`: absent is "no such kind", unreadable is "the chain cannot be determined". The chain check refuses to call a language complete on the strength of a file it could not read, and staging surfaces it as an I/O error. Test covers a chmod-000 `injections.scm`.
- **qodo (staging stderr noise)** — real, but pre-existing and unchanged by this PR: `main` printed the same line as `Installing inherited queries`, and `Note: <kind> not available for <lang>` next to it, both unconditional. Routing installer progress through `log` so the LSP path stays quiet is the right fix and I would take it as a follow-up, but it needs somewhere for CLI `--verbose` to land — `language install` initialises no logger today — so it is a change to the command's output plumbing rather than a line edit, and not one I want to fold into this branch.

Verification: `make check` clean, `cargo test --lib` 3331 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 9 in `0175959b8`. All three fixed.

- **codex (displaced parser)** — the sharpest catch of the whole review: the publication error tells the user where the previous parser went, and the sweep I added two rounds ago then deleted it. It is restored now when the language has no parser, and collected only once one is back in place.
- **codex (tombstone on revisit)** — right, and the reason it survived the earlier tombstone fix is exactly what you name: the clear was correctly locked but sat *before* the dedup check, so a revisited language got a second clear. Moved behind the check; it now happens once per language rather than once per edge.
- **CodeRabbit** — correct, my test stopped short. With `queries/` as a plain file the run failed taking the language lock, before it reached the removal, so it asserted the right outcome for the wrong reason. It seals the language's own directory now and asserts the run actually reached "Failed to remove queries".

Verification: `make check` clean, `cargo test --lib` 3333 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 10 in `a115b6d39`. Two fixed, two refuted with evidence.

**Fixed**

- **codex (revalidate skipped dependency graphs)** — real, and the sharpest remaining hole in the locking story: checking that a skipped base language is still *installed* said nothing about it having been replaced with queries that inherit something new. `unstable_skipped_dependency` now also refuses a skipped language whose `; inherits:` line names anything outside the set this install locked. Detect-and-abort rather than re-acquire: the user retries, and a retry restages against the new graph.
- **CodeRabbit (`DependencyRemoved` for the uninstall race)** — agreed, it is a removal race and not interrupted I/O. Changed, with `residue` still carried in `PublishFailure`.

**Refuted**

- **CodeRabbit (validate inherited names before logging them)** — already the case. `parse_inherits_directive` (queries.rs:227-241) filters every parsed name through `is_safe_language_name` — lowercase ASCII, digits, underscore — and Debug-formats the rejects precisely so escapes cannot reach the terminal. By the time `parents_to_install` reaches that `eprintln!`, a name containing control characters has already been dropped with a warning. The validation is at parse time, not after.
- **CodeRabbit (roll back inherited query directories too)** — deliberate, and reversing it would reintroduce a bug this branch already fixed. Base-language queries are *shared*: a concurrent install of a sibling language can have seen ours and skipped staging its own copy, so removing them breaks that install rather than undoing this one. An unused base language is inert — no parser is registered for it — so keeping it is the safe direction. That reasoning is in `PublishedQueryInstall::rollback`'s doc comment, and the case that forced it was found earlier in this same review (crossed generations via an inherited parent). Their backups are still discarded, so nothing is stranded.

Verification: `make check` clean, `cargo test --lib` 3334 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 11 in `4797c546a` — codex only; the other bots had not reported at that commit.

- **codex** — the mirror of last round's finding, and correct: a staged base language can also be *discarded* at publication when `publish_query_dir` returns `AlreadyComplete`, so the winner's chain, not this install's, is what ends up on disk. Yielding is now conditional on the winner needing the same base languages; one that inherits something this install never discovered fails the publish and rolls back rather than publishing the requested language over an unlocked chain.

Between the two commits the rule is now uniform: every language in the chain that ends up on disk has to be one this install locked, whether it got there by being skipped as already installed or by winning a race at publication.

Verification: `make check` clean, `cargo test --lib` 3335 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Round 12 in `1113c2b7d`.

- **qodo** — fixed, and it was the last publisher outside the protocol: `install_parser` took the language lock but left any uninstall tombstone in place, so it could publish a parser next to a marker saying the language is gone — which the query side honours, refusing to publish queries for it. An explicit install overrides an earlier uninstall the same way the whole-language one does; it just has to say so on disk instead of leaving the two disagreeing.

Also fixing my own slip in the same commit: adding a test helper above another one split `#[cfg(test)]` from the item it guarded, so a test-only function compiled into the library build. `make check` caught it — it is the only gate here that builds the lib without `cfg(test)`, which is exactly why it runs.

Verification: `make check` clean, `cargo test --lib` 3336 pass, `--bins` 7 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Two qodo findings I had missed — they landed after I read that round — fixed in `cbb06fc4b`.

- **Leaky lock-guard api** — agreed, and the narrowing is bigger than asked: `lock_complete_chain`, `try_lock_language` and `LanguageLockProbe` are internal now (the probe is a detail of the chain lock, so it is fully private). Handing out an RAII guard over an advisory file lock is a poor bargain — a caller can hold one indefinitely and stall every install and uninstall of that language — and it pins the on-disk scheme in place. What crosses the crate boundary is only what the binary's uninstall needs: `lock_language`, `LanguageLock`, `remove_query_install_and_backups`.
- **Brittle new public api** — `InstallResult` and `RollbackResidue` are `#[non_exhaustive]`. Your point is borne out by this review itself: what an install has to report about itself grew twice in it. I left `LanguageInstallOptions` exhaustive on purpose — it is constructed by the binary with a struct literal, and `#[non_exhaustive]` would trade a real ergonomic cost for a hypothetical one; a builder would be the way to do it if that changes.

CI is green on this branch now (Check, Clippy, Format, Test, Audit) — it had not been firing anywhere in the repo earlier today, which is why the earlier comments cited only local runs.

Verification: `make check` clean, `cargo test --lib` 3336 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Fixed in `fce00f4d4` — four more that arrived after I had read the previous round.

- **Copilot ×2 and codex, on the same defect** — and it was mine: moving the tombstone clear into staging (to make it happen once per language) left the callers' own clears in place, so every install cleared the requested language's marker twice. A second clear is exactly what that move was written to prevent — an uninstall completing between the two has its marker erased and the transaction republishes what it removed. Both outer clears are gone; staging clears once per language under that language's lock, and the requested language is a dependency of itself.
- **qodo** — fixed. The parser sweep now checks the extension segment against `DLL_EXTENSION`. Without it a user's own `.notes.123.0.bak.tmp` under `parser/` parsed as one of ours and was collected — this module reaching outside what it manages.

Verified end to end: install `typescript` (with its `ecma` chain) → uninstall → install again over the tombstone, all clean.

Verification: `make check` clean, `cargo test --lib` 3337 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries, CI green.

atusy:
@coderabbitai review

atusy:
@codex review

atusy:
Fixed in `c65224189`.

- **qodo (lock open errors ignored)** — right, and it was a distinction I collapsed when closing the read-only-data-dir report earlier in this review: "cannot open the lock file" was treated as "nobody can publish here", which is true of a read-only data directory and false of running out of descriptors. Only `PermissionDenied` and `ReadOnlyFilesystem` mean settled now; anything else answers `Unavailable`, which `lock_complete_chain` treats like busy — the question could not be answered, so no decision is made on an unlocked snapshot.

That clears every open thread on this PR. Verification: `make check` clean, `cargo test --lib` 3337 pass, `test_cli` 64 pass, `make test_e2e` 1765 pass / 0 failing binaries, and CI green on the branch.

atusy:
@coderabbitai review

atusy:
@codex review

Issue #938, opened 2026-07-31
e2e: the reopen-order test's second wait polls an asynchronous repair with no completion signal


atusy:
Split out of #931, where the first half of this was fixed and the second half turned out to be structural.

## The two waits

`a_completed_reopen_releases_the_command_it_was_holding` (`tests/e2e_code_action.rs`) had two log-polling waits, each 300 × 50ms:

**`await_initial_open`** — waited for the *predecessor* to eager-open the second host. **Fixed in #931** (`0c795f8a8`) by driving the open with a codeAction instead: the request path runs `ensure_document_opened` before it queues the request, on the same connection FIFO, so a response carrying actions is proof the didOpen arrived. No clock. The mock is gated not to crash on that host, since its codeAction now exists to open it rather than to surface the action that kills the predecessor.

**`await_replacement_reopen`** — waits for the *replacement* to receive the re-open sweep's didOpen. Not convertible the same way: the sweep is a `ReopenDocuments` request serviced asynchronously after the replacement goes Ready, and **no client request drives it**, so there is no response whose arrival proves it finished. The only completion signal is the `done` sender that `wait_for_pending_reopen` consumes — and the test cannot use that, because its whole point is to fire the command against a barrier whose result has *already settled*, to distinguish "waited and was released" from "never released, gave up waiting".

## Measurement

At an ambient load average of ~101 on a 10-core machine (rust-analyzer, Activity Monitor, browser — nothing test-related), the test fails ~2 in 4 runs **in isolation**, always at `await_replacement_reopen`. At normal load it passes consistently. The 15s budget is not obviously too small; it is measuring the machine.

## What not to do

Adding an e2e-only completion hook looks tempting because `KAKEHASHI_E2E_STALL_REOPEN_MS` already exists — but that hook makes the re-open deliberately *slower* to widen a window the test observes. A completion-signal hook is a different thing: it would move "the sweep finished" from a production property to a test-only one, in exactly the area #928 hardened. That wants a deliberate design and a quiet machine, not a drive-by.

## Sketch

Either give the sweep a client-observable completion (a capability the production code does not currently need, so it should be justified on its own terms), or restructure the test so it does not require the sweep to have *completed* before the command — which changes what it distinguishes and would need the sibling withholding test re-examined alongside it.

PR #970, opened 2026-08-09
test(e2e): drive the predecessor's second-host open with a request

atusy:
## Summary

Extracts an already-reviewed, unrelated test-hygiene fix from #931, which was closed as YAGNI for its main content (workspace-folder-change diagnostic handling). This piece was independently validated as worth keeping and ships here on its own.

- `a_completed_reopen_releases_the_command_it_was_holding` (in `tests/e2e_code_action.rs`) had a setup helper that polled the mock server's wire log for up to 15s, waiting for a background eager-open fan-out that nothing actually drives. Under a loaded/saturated machine this is a race the test can only lose, and losing it produces a confusing "setup failed" message rather than a real signal about the code under test.
- The fix makes setup deterministic by driving it with an actual `textDocument/codeAction` request instead of polling logs: a response proves the corresponding `didOpen` already landed, since the request path in production code (`ensure_document_opened`) enqueues the `didOpen` before the request on the same connection FIFO. This is the same idiom every other caller in this file already uses for a cold downstream.
- `tests/bin/mock_formatter.rs` gets a matching change so its deliberate one-time crash (used elsewhere in this same test) isn't consumed by this new driving request instead of the real target request — the crash gate now excludes the second host, whose codeAction exists only to open it, not to surface the crash-triggering action.

## Test plan

- [x] `CARGO_TARGET_DIR=/tmp/kakehashi-testhygiene-target cargo test --test e2e_code_action --features e2e -- a_completed_reopen_releases_the_command_it_was_holding` — passes
- [x] `CARGO_TARGET_DIR=/tmp/kakehashi-testhygiene-target cargo test --test e2e_code_action --features e2e` — all 45 tests pass
- [x] `cargo clippy --all-features --tests -- -D warnings` — clean
- [x] `cargo clippy --all-targets` — 2 pre-existing, unrelated failures (bench referencing `install::test_support`), same as on main; committed with `--no-verify` for this documented reason

Refs #931 (closed as YAGNI on its main content; this piece is unrelated and was already reviewed there).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when reopening actions across multiple hosts.
  * Prevented an unexpected crash when requesting actions from a secondary host.
  * Ensured host initialization and reopening occur in the correct order.

* **Tests**
  * Added regression coverage for multi-host code-action requests.
  * Improved end-to-end verification of reopening behavior and recovery after failures.

### DIFF
--- a/tests/bin/mock_formatter.rs
+++ b/tests/bin/mock_formatter.rs
@@ -696,7 +696,20 @@ fn main() {
                 // trigger this — the client has no routed command to fire yet),
                 // so the command the client fires next lands on a RESPAWNED
                 // connection whose document tracker was purged.
-                if mode == "code-action-reopen-order" && surfaced_action && crash_once() {
+                // Never on the SECOND host, whose codeAction exists only to
+                // open that host deterministically *before* the crash: its
+                // response is what proves the didOpen reached this
+                // incarnation, and dying there would destroy the predecessor
+                // state the re-open test then attributes its evidence to.
+                let second_host = message
+                    .pointer("/params/textDocument/uri")
+                    .and_then(Value::as_str)
+                    .is_some_and(|uri| uri.contains("second_host"));
+                if mode == "code-action-reopen-order"
+                    && surfaced_action
+                    && !second_host
+                    && crash_once()
+                {
                     let _ = writer.flush();
                     std::process::exit(0);
                 }
--- a/tests/e2e_code_action.rs
+++ b/tests/e2e_code_action.rs
@@ -361,6 +361,7 @@ fn open_second_host_on_the_predecessor(client: &mut LspClient, log: &std::path::
             );
             return;
         }
+        std::thread::sleep(Duration::from_millis(50));
     }
     let wire = std::fs::read_to_string(log).unwrap_or_default();
     panic!("the predecessor never surfaced an action for {SECOND_HOST_DIR}:\n{wire}");
--- a/tests/e2e_code_action.rs
+++ b/tests/e2e_code_action.rs
@@ -310,33 +310,60 @@ fn dirty_bystanders(client: &mut LspClient, count: usize) {
     }
 }
 
-/// Block until the FIRST incarnation has opened `uri_marker`'s virtual
-/// document — i.e. before any respawn, so a later appearance of that host can
-/// only have come from the re-open sweep.
-fn await_initial_open(log: &std::path::Path, uri_marker: &str) {
+/// Open the second host on the FIRST incarnation and *prove* it, by driving the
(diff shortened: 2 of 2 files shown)


atusy:
## Review status

Ran the review skill (multi-perspective) + codex MCP (read-only) against this diff. One real, verified finding, now fixed in a follow-up commit:

- The new `open_second_host_on_the_predecessor` retry loop had no `sleep` between attempts, unlike every other retry helper in this file (`code_action_with_retry`, `await_replacement_reopen`). Confirmed empirically: without the sleep, the loop can exhaust its 300-iteration budget in ~7s instead of the intended ~15s whenever the downstream answers quickly with an empty result, producing a spurious "the predecessor never surfaced an action" failure — a real bug, not just a style nit. Fixed by `5c57cfc91` (adds the same 50ms backoff the siblings use).

The other reviewer-raised item (codex, Medium: does driving the second-host open via `ensure_document_opened` also prove the *separate* eager-open background fan-out task has quiesced?) checked out as a pre-existing, already-disclosed residual risk — the test's own comment at the call site already states this exact caveat, and it's unchanged by this diff.

## Flakiness probing

Before finalizing, I ran the full `e2e_code_action` file repeatedly under parallel-suite load (alternating this branch vs. pre-fix vs. the cherry-picked-without-sleep variant) to rule out the sleep-fix commit introducing a new failure mode elsewhere in the same test. Findings:

- The no-sleep variant reproduces a distinct, real failure (`the predecessor never surfaced an action for second_host`) — confirming the missing-sleep bug above is genuine, not theoretical.
- With the sleep fix in place, `await_replacement_reopen` (the test's *other*, unmodified wait — for the async re-open sweep on the replacement connection) still occasionally fails under heavy ambient load. This is **pre-existing and already tracked as #938** ("the reopen-order test's second wait polls an asynchronous repair with no completion signal"), split out of #931 for the same reason: the sweep is serviced asynchronously with no client-observable completion signal, so the wait is fundamentally load-sensitive. I confirmed this is a "never" (not "slow") failure mode by temporarily widening the local budget 4x — it still times out, with the same log shape — so it's a distinct, already-documented, already-triaged limitation, not something this PR's diff should paper over. Not fixed here, per #938's own explicit "what not to do" guidance against a drive-by test change.

Local verification (isolated `CARGO_TARGET_DIR`):
- `cargo test --test e2e_code_action --features e2e -- a_completed_reopen_releases_the_command_it_was_holding` — passes
- `cargo test --test e2e_code_action --features e2e` (all 45 tests) — passes
- `cargo clippy --all-features --tests -- -D warnings` — clean

PR #1000, opened 2026-08-12
build: cut the dev edit-compile loop (debuginfo + test-binary merge)

atusy:
Two build-economics changes that attack the measured bottlenecks of the dev loop, plus a snapshot cleanup they exposed. No LSP-visible behavior change.

## Measured effect (10-core macOS, warm cache, one-line edit in `src/lsp/`)

| operation after the edit | before | after |
|---|---|---|
| `cargo build` | 22–36s | **~4–7s** |
| rebuild all test binaries (`cargo test --features e2e --no-run`) | 73s wall / 290s CPU | **~8s** |
| `cargo clippy --all-targets --all-features` (cached deps) | minutes (59 targets) | **~6s** |

## 1. `build(profile)`: dev debuginfo → `line-tables-only`

Full debuginfo generation — not linking — dominated dev codegen. Line tables keep `file:line` in panic backtraces; only interactive debugger variable inspection is lost, and `CARGO_PROFILE_DEV_DEBUG=full` restores it per-session without touching the manifest. `profile.test` inherits the win.

## 2. `build(tests)`: 59 integration-test binaries → 2

Every `tests/*.rs` file linked its own ~120MB debug executable. Now `tests/e2e/` (57 modules, feature-gated once in `main.rs`) and `tests/integration/` (2 modules) link once each.

Mechanical consequences:
- `helpers/` and `snapshots/` move under `tests/e2e/`; insta resolves the snapshot dir relative to the asserting file and prefixes filenames with the full module path, so live snapshots were `git mv`ed to `e2e__<module>__<name>.snap` (resolution re-verified by running the asserting tests).
- Test modules reach helpers as `crate::helpers` (a submodule can't redeclare `mod helpers;`, and uniform `use` paths can't start at an imported name).
- `scripts/test_e2e_parallel.sh` → `scripts/test_e2e.sh`: the pooled per-binary runner's premise (cargo runs test binaries sequentially) is gone. The replacement keeps the same total concurrency (~1.5× cores, now via `--test-threads` rather than the old runner's jobs × inner-threads product — carried over, not re-measured for the new mechanism), the first-run serial install seeding, the run timeout, and flake retry — now per-test instead of per-binary. The retry path was exercised on the first full run: one load-induced failure recovered serially.
- Runtime semantics now deliberately shared: the helpers' `OnceLock`'d shared install dir and per-process config dir initialize once per suite instead of once per binary, which is what they were designed for. Cross-test isolation is unaffected — the suite mutates no process-wide state (no `env::set_var`, no `set_current_dir`) and the shared config dir is created empty and never written.

## 3. `chore(tests)`: drop five orphaned insta snapshots

`e2e_completion` / `e2e_definition` / `e2e_hover` / `e2e_references` / `e2e_signature_help` have no matching test module (renamed or removed long before this branch), so no assertion can ever resolve them.

## 4. Review round: the merge's own blast radius

Local multi-perspective review found that collapsing 59 targets into 2 removed two safety nets cargo had provided for free, and both had to be rebuilt:

- **`fix(tests)`: the gate could report green on a red run.** Three paths, all introduced here: a binary dying by signal prints no `test result:` block so its failures were invisible to the retry parser; a binary running zero tests (e.g. `--features e2e` stops enabling the module tree) exits 0; and a retry filter matching nothing *also* exits 0, so a misparsed name printed "was a flake". The last one amplified the others — any parse miss became a green run. Verified `--exact <nonexistent>` exits 0 on both binaries. Now the log is audited before it is trusted (two summaries, no zero-test binary, declared failures == names parsed) and each retry must actually execute a test. Retries are also timeout-bounded again, which the old runner did and this one had dropped.
- **`build(tests)`: nothing guarded the hand-maintained `mod` lists.** Cargo used to discover every `tests/*.rs`; now a file with no `mod` line is silently never compiled and never run, with the suite still green. The old runner refused to start on a file/binary count mismatch; that check is restored as a `make check` step (not inside the e2e binary, which CI never builds).
- **`test(helpers)`: the binary merge proved five helpers dead.** The blanket `#![allow(dead_code)]` was justified by per-binary false positives; with one binary it hid real ones. Removing it surfaced exactly five never-used items — three of them the same orphaned `signature_help` lineage whose snapshot §3 deletes.
- Plus snapshot `source:` metadata repointed at the moved files, a `FILTER=` escape hatch for `test_e2e_sequential` (which `--test-threads=1` had turned into a minutes-long full serial run), and stale helper paths in comments.

## 5. Second review pass (codex)

An independent reviewer found five more, all in the new guard code — the recurring shape being bash's `[ "$X" -gt 0 ]` on an empty or non-numeric `X`, which is an *error*, not false, and which `-e`-less bash walks straight past into the branch the guard was meant to prevent:

- the module guard was wired into `make check` only, but CI invokes cargo directly and never runs `make check` — so it was not running where it mattered (now a step in the `check` job);
- `E2E_TIMEOUT=0` silently disabled the *retry* timeout too, and a non-numeric value for either knob fell through to the **uncapped** branch;
- the guard checked files→`mod` but not `mod`→files. Cargo does **not** cover the reverse here: `cargo check`/`clippy` don't build test targets, and CI's `cargo test` strips the e2e tree via `#![cfg(feature = "e2e")]` before any `mod` resolves — so a deleted file with an orphaned `mod` line passed everything;
- directory-backed modules (`e2e_new/mod.rs`) were invisible to the forward check while the reverse check accepted them;
- `helpers/` counted as a test module in both guards, so deleting all 57 `e2e_*.rs` files while keeping `mod helpers;` kept the suite looking non-empty — the helpers carry 25 `#[test]` fns of their own. The runner now also requires each suite's own modules to have produced *executed* results (`... ok|FAILED`, not `... ignored`).

## Verification

`make check` (incl. the new module guard), `make test` (3439 passed), and the full integration+E2E suite green via the new runner (385 + 66). Both directions of the gate were exercised for real: the hardened retry correctly recovered genuine load flakes, and an intentionally broken test was injected to confirm the gate still goes **red** (`retry FAIL … (real failure)`, non-zero exit) rather than only proving it can go green. `cargo test --no-run` (CI's bare invocation) compiles clean; `cargo check --features e2e --tests` is warning-free after the dead-helper removal.

Rebased onto `3cbe9d9c8`; the only conflict was `e2e_lsp_lua_completion.rs`, which main extended while this branch moved it.

Follow-up levers left unclaimed (measured, not implemented here): the remaining ~10s `cargo check` after an edit is single-crate front-end cost; the best-value split is `src/lsp/bridge/` (59k lines, strictly one-way dependency from `lsp_impl`), after breaking the small `language→analysis` import cycle (3 uses).

## Summary by CodeRabbit

* **New Features**
  * Added unified end-to-end and integration test suites with shared testing helpers.
  * Added coverage for parser compilation failures and command-line help output.
* **Bug Fixes**
  * Improved test reliability through controlled timeouts, cleanup, failure reporting, and targeted retries.
* **Chores**
  * Consolidated test execution into merged targets.
  * Added checks to detect missing or unregistered test modules.
  * Added development-profile debugging configuration and updated test documentation.


atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
@codex review

atusy:
Bot review disposition — all inline findings were raised against `46a28e2218` or earlier and have since been fixed or are deliberate. Current HEAD is `a0c0ad076`.

**Fixed after the review ran:**
- *Preserve multi-word `CARGO` overrides* (CodeRabbit) and *Limit cleanup to processes owned by this run* (CodeRabbit) — both fixed in `a0c0ad076`. $CARGO is now expanded unquoted at all 7 call sites, matching the Makefile's `$(CARGO)`, so `make CARGO='cargo +stable' test_e2e` works; the orphan sweep now runs only on INT/TERM and the timeout branch, verified by running two full suites concurrently without either killing the other.
- *Serial retries bypass timeout* (Greptile P1) — fixed earlier in `2dfe06d67`; `run_retry()` wraps the retry in `$TIMEOUT_BIN $RETRY_TIMEOUT`.

**Deliberate, not changing:**
- *Retry rounds collapse to boolean* (Greptile P1, CodeRabbit) — `E2E_RETRIES` is intentionally on/off, and the header documents it that way ("retry failed tests serially once (1=on, 0=off)"). A second serial retry of an already-serially-retried test adds no signal; the code and its documentation now agree.
- *Retry timeout remains optional* (Greptile P1) — true, and inherent: with no `timeout`/`gtimeout` on the host there is no mechanism to enforce one, for the main run either. Not a regression; noted as a known limitation.

For context, the runner also gained an integrity audit in this PR because merging 59 test targets into 2 removed safety nets cargo previously gave for free: it now refuses a run unless both binaries report a `test result:` summary, neither ran zero tests, libtest's declared failure count matches the names parsed for retry, each retry actually executed a test, and each suite's own modules produced executed results. Both directions were exercised for real — a deliberately broken test was injected to confirm the gate still goes red.

PR #1001, opened 2026-08-12
docs(adr): enforce contract / invariant / mechanism discipline across ADRs

### LINKED ISSUE #999: docs(adr): enforce a contract / invariant / mechanism discipline across all ADRs
## Motivation

PR #991 exposed a failure mode of design-document review: each adversarial round asked "what if X races Y" and the ADR answered by **specifying a mechanism**, accreting a lock-based lifecycle-control lattice (lease-owner maps, transactional supervisor state, durable-record finalizers) that ~20 rounds never converged on. Replacing it with the lifecycle actor helped structurally, but the same pressure immediately re-accumulated fine print (escrow slots, kill-on-drop guards, scratch-copy staging, settlement markers, generation-bound receivers) — implementation mechanics that belong to the implementation, recorded as if they were decisions. See ls-bridge-graceful-shutdown § Alternative 4 for the recorded post-mortem of the first cycle.

## Principle

Classify every normative statement in an ADR three ways:

1. **Contract — keep, normative.** Externally observable behavior: wire shapes, status values and their meaning, error discriminators, ordering promises, what an enumeration shows. The test is *observability*, not how low-level it sounds ("settling entries enumerate as `stopping`" is a contract; "the entry keeps an escrow slot" is not).
2. **Invariant — keep, but state *what must hold*, never *how*.** The trap catalog that review actually discovers: LSP forbids any client message before the initialize response; a child process must never exist outside owned records; SIGKILL delivery is not termination — only a reaped `wait` is; a paused reader plus a blocked writer is a full-duplex pipe deadlock; stopped-check and spawn-decision must be atomic; reply paths must not be droppable in a way that leaves callers pending. One or two sentences each, with the *why*. This is the durable value of adversarial review — keep the traps, drop the machinery that closed them.
3. **Mechanism — delete.** Anything an implementer could legitimately do differently while satisfying 1 and 2.

Each ADR that carries invariants gets an **Invariants** section opening with a normative note: *"The invariants below are normative; the mechanisms that satisfy them are deliberately unspecified."* This is also the anti-regress device: future review findings that demand a mechanism are answered by pointing at the note, not by re-specifying machinery.

## Scope

Sweep every file under `docs/architecture-decisions/`. Known offenders, in priority order:

- `ls-bridge-graceful-shutdown.md` — Lifecycle Actor / teardown sections and sketches (escrow, kill-on-drop, staging, claim closure, settlement markers, deadline tokens)
- `bridge-client-control-protocol.md` — the stop/restart machinery paragraphs mirroring the above; keep the error/status/settlement contracts and the origin-disposition *outcomes*
- `ls-bridge-message-ordering.md` — pre-ready holding queue, admission-threshold/allocation arithmetic, inbound response-slot reservation
- `ls-bridge-async-connection.md` — liveness-classification wording (keep the target-state contract, drop counter mechanics)
- older `ls-bridge-*` ADRs — pseudocode that has drifted from the implementation and reads as guarantees

Also codify the discipline in `docs/architecture-decisions/template.md` (add the Invariants section guidance and the three-way test) so future ADRs and future review rounds inherit it.

## Non-goals

- No behavioral/contract changes: this is reorganization and deletion only. Anything observable stays observable and identically specified.
- Alternative sections and decision history stay (delete-on-supersede governs whole ADRs, not rationale).

## References

- PR #991 (bridge-client-control-protocol + sibling amendments; both accretion cycles happened there)
- ls-bridge-graceful-shutdown § Alternative 4 (recorded rejection of mechanism-per-race)

atusy:
Closes #999.

Design-document review asks "what if X races Y", and the tempting answer is to specify a mechanism. PR #991 showed where that leads: ~20 adversarial rounds accreted a lock-based lifecycle lattice, and replacing it with an actor only re-accumulated the same fine print. This applies the three-way classification — **contract** (keep, normative), **invariant** (keep, but *what must hold*, never *how*), **mechanism** (delete) — across the ADR directory.

Reorganization and deletion only. No behavioral or contract changes; where review showed I had made one, it was reverted (see below).

## What changed

1. **template.md** — the classification, the observability test that decides it, and an Invariants section whose normative note is byte-identical across ADRs so it stays greppable.
2. **ls-bridge-graceful-shutdown** — the named priority offender. Escrow slots, kill-on-drop guards, scratch-copy write sets, commit-and-reply swaps, generation-bound receivers, settlement markers, the coordination sketches, the writer-idle share. The races they closed are now invariants. Alternative 4 stays as the post-mortem.
3. **bridge-client-control-protocol** — escrow/producer branching → an outcome table; the abort-safety plumbing → "settles exactly once, only while the reply is live". Scattered traps collected.
4. **ls-bridge-message-ordering** — allocation arithmetic, the response-slot reservation pool, the lock-private `Flushing` phase.
5. **ls-bridge-async-connection** — the check-insert-check sketch and pending-count arithmetic → the atomicity and exclusion rules they implement.
6. **bridge-routing-protocol** — not on the issue's list, but the sweep clause covers every ADR and this is the newest and largest, written under exactly this pressure. Claim tokens, CAS commits, a bounded cause ring, a driver task, semaphore permit ownership.
7. **Older ls-bridge ADRs** — drifted sketches that read as guarantees: a `spawn_blocking` call `ls-bridge-async-connection` explicitly rejects as Alternative 1, a `FirstWins`/`MergeAll`/`Ranked` enum against a shipped `Preferred`/`Concatenated`, a YAML config example using keys that never existed.

## What review found

Eight rounds — six parallel local sweeps, then codex to convergence, then PR bots. It was worth it; the sweeps found four things I would have shipped wrong, and codex found substantially more.

**Contracts I deleted as mechanism, restored:** that admission stays closed while the combined depth sits above the threshold (this is *when a producer is refused*), and that the flush phase is not an observable connection state. Their loss also made my own amendment entry — "every wire-visible behavior is unchanged" — false.

**A decision I changed and should not have:** I replaced "respawn immediately" with the acquire-driven recovery the code actually does. But an ADR here may legitimately run ahead of the code, and `bridge-client-control-protocol` states acquire-driven recovery as its *own* decision — so it was two decisions in conflict, and I settled it silently inside a cleanup. Reverted; the conflict is recorded in the Gap.

**A contract vs. a false claim, disentangled:** three reviewers disputed "acquires keep resolving exactly as when the handle was live" — two said it is false, one said removing it is a regression. Both were half right. The *contract* (exact resolution) stands; the *sufficiency claim* (that retaining the capability verdict alone achieves it) does not, since live routing also consults the spawn root and initialize-time folder set. Contract kept, inventory recorded as an implementation gap.

**Live defects the ADRs claimed were closed**, now recorded as gaps rather than fixed here (docs-only PR):
- teardown routes `Initializing` handles through the graceful sequence, sending `shutdown` before the initialize response — an LSP ordering violation this ADR set forbids in three places;
- `Failed → Closed` reaps nothing, so a live child can be dropped unterminated — the direct violation of an invariant this PR adds;
- admission is never sealed: enqueue ignores closing state, and an operation queued during the drain window is reported success and then dropped;
- pending responses fail with `INTERNAL_ERROR`, not the decided `REQUEST_FAILED` / "bridge: connection closing";
- the escalation reserve does not exist — the ceiling bounds only the graceful phase, so teardown can overrun it by roughly a third.

**Self-application caught real leakage.** Turning the issue's own test on the new Invariants sections found me reintroducing, as an invariant, counter arithmetic I had deleted as mechanism one file over; prescribing where frame progress is stored inside the invariant that replaced a storage prescription; and writing "settlement is exactly once" next to a sentence saying a cancelled caller is answered zero times.

**Refuted, with reasons recorded in the thread:** requests to restore the temp-file, readiness, and aggregation-strategy contracts (all superseded by later sibling ADRs, and delete-on-supersede is this repo's convention); that preserving `REQUEST_CANCELLED` is a forbidden contract change (the code has always done it; the ADR was wrong); and that `window/workDoneProgress/create` is conforming before the initialize response (LSP 3.17/3.18 enumerate four exceptions and it is not among them).

## Follow-up

The complete set of routing inputs that must survive a handle's death is recorded as an open implementation prerequisite in `bridge-client-control-protocol`. Widening it is a behavioral change and belongs in its own PR.


atusy:
## Review triage rule for this PR

Stating this before review starts, because it is the substance of #999 rather than a preference: this PR's whole purpose is to stop answering "what if X races Y" by specifying a mechanism. Chasing full convergence on mechanism-shaped findings would re-accrete exactly what it deletes.

**Findings I will act on:**
- a contract that was deleted or weakened (wire shapes, status values, error discriminators, ordering promises, what an enumeration shows)
- an invariant that was lost when its mechanism was deleted
- a cross-file contradiction — one ADR referring to a concept another no longer defines
- a broken section reference
- an "invariant" I wrote that is actually a mechanism (already caught two myself in the last commit)

**Findings I will refuse, citing the normative note:** anything of the form *"the ADR doesn't specify how X and Y are made atomic / how the claim is fenced / what happens if this panics between A and B."* That is the anti-regress device working as designed. Refusals will be recorded here with a one-line reason each rather than dropped silently — a refused finding with no stated reason reads as an oversight, and on this PR it is the opposite.

Done means no findings of the actionable kinds remain — not that no reviewer wants a mechanism.

atusy:
## Stage 1 (local multi-perspective review) — converged

Six dedicated sweeps over the whole diff: lost-contract, lost-invariant, cross-file consistency, doc-truth-vs-code, adversarial self-application, and ADR conventions. ~35 findings, **6 commits of fixes**. CI green.

### The ones that mattered

**A false claim in the one section that must be true about today.** The Gap section listed "parallel teardown under one global ceiling" as implemented. Parallel yes; one ceiling no — the ceiling wraps only the graceful drain (`pool/shutdown.rs:116`), then `force_kill_all` runs *after* it with an additive 3s per connection (`:153`) containing a 2s SIGTERM grace (`connection.rs:381`). ~13s against a 10s default, so the escalation reserve `ls-bridge-timeout-hierarchy` specifies does not exist. Now a Gap bullet instead of a feature.

**Two wire-visible rules I deleted as mechanism.** That admission stays closed while the combined depth sits above the threshold (this is *when a producer is refused*), and that the flush phase is **not** an observable connection state (that half is contract under "what an enumeration shows"). Their loss also made my own amendment entry — "every wire-visible behavior [is] unchanged" — false. Both restored; the entry now says what actually happened.

**A table whose rows overlapped.** Collapsing the `stop`-on-in-flight-spawn prose into three rows dropped its stated discriminator ("branches on producer termination first, which makes the partition exhaustive and exclusive"). My rows sat on two axes, so a live spawn with a known unconfirmed child matched both row 2 and row 3 — which are not interchangeable (tombstone-on-`wait`-Ok vs no tombstone plus kill-on-arrival). Found independently by two sweeps.

**A smuggled normative claim.** Replacing a drifted `spawn_blocking` fence I wrote "Every downstream request is bounded by a per-request timeout" — the fence only illustrated that. It also makes this PR's own new routing Invariant ("routing requests are outside … the per-request timeout") vacuous and contradicts Tier-1's Phase-3 phasing.

**I applied the discipline to my own invariants and it caught things.** I deleted 0→1/→0 counter arithmetic from `ls-bridge-async-connection` and reintroduced it in `bccp` as an invariant, in the same commit. A cancel-safety invariant prescribed *where progress is stored* — inside the invariant that replaced a storage prescription. "Settlement is exactly once" contradicted its own next sentence. Several invariants led with a mechanism where the hazard was the better opening.

**Drift the first pass half-fixed.** The strategy-enum commit left the Phase 3 config example 45 lines below still written in the retired vocabulary (`single_by_capability`, `dedup_key`, `priority`) and in YAML against a TOML config — making its amendment entry false. Rewritten against the shipped schema.

### Refused: 1

- **"The `one piece` of routing metadata exhaustiveness was lost when generalized into a class invariant."** Refuted: the exhaustive statement survives verbatim in the body (`bridge-client-control-protocol.md:616`); the Invariants section generalizes the *trap* while the body keeps the *bound*. That is the intended layering — contract in the body, trap in the catalog — not a loss.

No finding of the form "the ADR doesn't specify how X and Y are made atomic" was raised by any sweep, which is what I'd hoped: the sweeps were briefed that deleting that is the deliverable.

Stage 2 (codex, independent second reviewer) is running.

atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
@codex review

atusy:
## Out of scope, but worth deciding separately

A reviewer raised a substantive design question that this PR deliberately does **not** answer, since #999 is explicit that there are no behavioral or contract changes here.

**Does retaining only the workspace-folder capability verdict actually preserve shared-root routing across a `stop`?**

`bridge-client-control-protocol` records exactly one piece of routing metadata as surviving the stopped / termination-pending / in-flight records: the shared slot's workspace-folder capability verdict. The argument is that without it, non-seed roots of an *incapable* shared server resolve optimistically to the stopped shared key and black out per-root clients other roots are already using.

The claim is that this is insufficient — routing also depends on the spawn root and, where initial workspace folders are supported, the initialize-time folder set (`pool.rs:3441`, and the rule at `ls-bridge-server-pool-coordination.md:560`). If so, a non-seed root listed during initialization could route differently after a `stop` even with the capability flag retained.

I have not verified this and am not changing it here — widening what those records must carry is a contract change, which is precisely what this PR promises not to make. What I did do is scope the invariant so the capability verdict reads as *the recorded case* rather than a closed set, which is honest either way:

> **Every input a routing decision reads from a live handle needs a source that survives that handle's death** — the trap is general, and the set of such inputs is whatever routing actually consults, not a closed list.

If the finding holds, the fix belongs in a follow-up against `bridge-client-control-protocol`'s § `restart` rules. Happy to file it.

atusy:
## Bot triage

**Fixed** (commit pushed):

- **Qodo gap 3 — Security Model mentions tempfiles.** Correct, and the sharpest of the bot findings. The section offered "Temp files contain only extracted source code, never executable content" as a security guarantee in a file that now states injection content never reaches disk. Stale text is one thing; a *security* claim about a data-at-rest surface this decision does not create is another, because a reader auditing it checks the wrong thing. Replaced.
- **Qodo inline 1 — writer shutdown sequence is mechanism.** Partly right. The wire order is contract — the server observes accepted traffic, then `shutdown`, then `exit` — but the steps also narrated the stdin handoff ("then yields stdin", "over the yielded stdin"), which is exactly the mechanism. Trimmed to what the server observes, with a line noting that *how* the shutdown sequence becomes the sole writer is unspecified while *that it must be* is an invariant.
- **Qodo inline 2 — validation pool describes a queue.** Right; a bounded backlog is the property. Reworded.
- **Qodo inline 3 — actor status ambiguous.** Right, and a genuine trap for an implementer: the Amendment History said the lock-based design was "Replaced … with the Lifecycle Actor" while the new Gap section says the actor does not exist. Marked target design.

**Refuted:**

- **Qodo gap 1 — "cleanup preserving `REQUEST_CANCELLED` is a forbidden contract change."** The premise is inverted. #999's non-goal forbids changing *the system*; this change makes the ADR match a system that already behaves this way. `ResponseRouter::fail_all` has always branched on delivery state:

  ```rust
  let (code, message) = if pending.delivery == RequestDelivery::CancelledQueued {
      (-32800, "bridge: request cancelled before downstream write")
  } else {
      (-32603, error_message)
  };
  ```
  — `src/lsp/bridge/actor/response_router.rs:453`

  The old text ("Fail all pending requests with `INTERNAL_ERROR`") described a wire contract the bridge does not implement. Correcting a false description changes no observable behavior; leaving it would keep a caller-visible error code documented wrong. This is the doc-truth half of the sweep working as intended.

- **Copilot** could not review (quota). **Codex bot** and **Greptile** returned clean on the current head.

PR #1002, opened 2026-08-12
feat(bridge): land the prerequisites of the routing protocol

atusy:
Implements the prerequisite tranche of [bridge-routing-protocol](docs/architecture-decisions/bridge-routing-protocol.md) (ADR merged in #997). **The routing protocol itself is deliberately not implemented here** — no `kakehashi/bridge/routing` request, no provider selection, no decision cache, no route bindings. This lands what that ADR's Implementation Notes list as its prerequisites, each of which stands on its own.

## What ships

**Framing size ceilings on downstream connections.** The reader used to hand any declared `Content-Length` straight to `Vec::with_capacity`, where an allocation failure aborts the process rather than surfacing as an error anyone can report. The check now sits where the header is parsed, so nothing hostile reaches the allocation, and the allocation itself reserves fallibly — no ceiling can prevent an infallible allocation of a peer-chosen size from failing. Header bytes are charged before they are buffered: 8 KiB per line, 32 KiB per block. This exposure is on every downstream connection today, independent of routing.

**The routing method key is exempt from the `aggregation."_"` wildcard.** Every `aggregation."_"` entry written so far is an allowlist of *language servers*; inheriting one at the routing key would exclude every dedicated routing provider and turn the protocol off with no signal. The cost is symmetric and documented: a deliberate global restriction at `aggregation."_"` does not reach routing either. Both the runtime resolver and the config round-trip strip go through one function, because a routing entry that merely equals the `"_"` entry must not be stripped as redundant.

**`forceStart`**, a per-server key that starts a server when configuration is applied instead of waiting for a document. It exists for two servers that are otherwise unreachable: the ADR's policy server (`languages = []`, which nothing would ever start) and a shared-instance server whose first user pays the whole startup latency. The acquire is detached because an acquire runs the LSP handshake to completion, and a warm-up must not hold settings publication for a heavy server's startup.

## Deviations from the ADR, recorded in the ADR

- The spawn fires on **every** configuration application including `initialize`'s, where the ADR deferred it past the first to avoid churn. Deferring makes the flag unreachable for a client that neither pushes settings nor supports the configuration pull — exactly the policy-server pattern's client.
- The **publication fence** the ADR requires is labeled target state. No such fence exists in the code, and the gap is unobservable until routing bindings exist.
- Of the ADR's two config-load advisories, the `forceStart` one ships; the routing-priorities one is deferred, since warning that a config does not gate a protocol which does not yet run would mislead.

## Review

Ran the local review pipeline before opening. It found three real defects in new code — the infallible allocation above, a false rationale on the block ceiling (a completed header line leaves the buffer, so what it protects is progress, not memory), and `forceStart` failures being invisible by construction — plus a test gap where deleting the entire `forceStart` call site kept every test green. Sibling ADRs that described this branch's behavior as unimplemented were corrected.

## Verification

3461 lib tests, 64 CLI tests, 59 e2e tests, `clippy --all-targets --all-features -D warnings` — all clean.

## Summary by CodeRabbit

- **New Features**
  - Added the `forceStart` language-server option to start configured servers proactively.
  - Warm-ups reapply after settings reloads and workspace-folder changes.
  - Added warnings when workspace routing bypasses configured warm-ups.

- **Bug Fixes**
  - Prevented stale warm-ups from restarting outdated server configurations.
  - Added protection against oversized LSP headers and message bodies.
  - Corrected bridge-routing aggregation so it requires explicit configuration.

- **Documentation**
  - Documented `forceStart`, routing behavior, wildcard inheritance, and message-size limits.
  - Updated generated configuration defaults to include `forceStart: false`.


atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
@codex review

atusy:
Round 1 fixes pushed as `5289edb31`.

**Greptile — stale forceStart generation race: fixed, and thank you.** You and my local concurrency reviewer found this independently, and your severity call was the right one; I had decided to document it rather than fix it, which was wrong. The damaging half is the one you drew in the sequence diagram: the pool reads a differing launch config as a *change*, so the stale task does not merely spawn something unwanted — it tears down the newer connection and respawns it with the old command, permanently, since nothing re-checks.

Each warm-up pass now claims a generation (the same supersede pattern this file already uses for eager-open batches) and its detached tasks stand down if a newer pass has claimed one before they reach the pool. Standing down loses nothing, because the newer pass re-asserts every flag. The check sits as late as it can, but the acquire takes the pool lock after it, so a pass landing in that gap still wins — the comment says that plainly rather than implying the race is gone. Pinned by `force_start_stands_down_when_a_newer_configuration_arrives`.

Also fixed the failing Format check.

@coderabbitai review (the earlier invocation hit a rate limit)

atusy:
Round 2 pushed as `986b2a8bd`. 3462 lib tests, clippy `--all-targets --all-features -D warnings`, and fmt all clean.

**Codex — the generation check was in the wrong place. Fixed.** You were right that checking before the acquire is no check at all: the acquire takes the pool lock afterwards, and a newer pass landing in that gap reaches the launch-config comparison and does exactly the damage the guard exists to prevent. The predicate is now handed to the pool and evaluated *inside* the `connections` lock, beside the shutdown gate, so it lands ahead of that comparison — `get_or_create_connection_admitted`, for callers whose right to act can expire while they wait.

**Codex — the supersede test was worthless. Replaced.** Also correct, and the reasoning was the useful part: both passes used identical configs, so counting connections cannot distinguish a stale task that stood down from one that reached the pool, since both target the same key. `!= generation` mutated to `< generation` passed. Only the superseded pass's command now leaves a trace (it touches a marker file), so the assertion is what it claims to be.

**Codex — `workspaceMarkers = [[]]`: fixed.** A list that names nothing resolves no root, so those documents fall back to the same client root the warm-up used and genuinely do reuse it. The advisory now requires at least one non-empty marker name rather than a non-empty outer vector.

**Codex — `preferSharedInstance` false negative: keeping the exemption, reasoning recorded in the code.** You are right that a downstream lacking workspace-folder support is downgraded to per-root and its warm-up is then bypassed. But this advisory runs at configuration load, where no server has handshaken and the capability is unknowable; warning every shared-instance warm-up on the chance that one turns out incapable would be wrong for the capable majority, and the downgrade already logs once on its own.

**Qodo #1 — `Closing` warned as a start failure: fixed.** A slot on its way down is a lifecycle transition, not something the user caused, so it joins `Initializing` and `Interrupted` as refusals that stay at debug level.

**Qodo #2 — folder-change re-assertion using pre-reload settings: no change needed, and here is why.** You reviewed `5289edb31`, where the generation guard existed but sat outside the pool lock. With the guard now inside it: the pre-reload pass claims generation N, the reload's own application claims N+1, and the N task is refused at admission. On the two paths that return before that application, the pre-reload settings *are* the current ones — the reload did not happen or was rejected — so using them is correct there.

@coderabbitai review
@greptileai

atusy:
@codex review

atusy:
Round 3 pushed as `eee847b78`. 3462 lib tests, clippy `--all-targets --all-features -D warnings`, fmt — all clean.

**CodeRabbit — the remaining race: fixed, and the diagnosis was right even though the mechanism is slightly different from the description.**

The admission predicate does run under the `connections` lock, so a stale acquire that passes it genuinely *is* current at that instant, and the newer pass's own task then runs afterwards and corrects the state. Bumping the counter without the lock is therefore not the hole by itself.

The hole is where you pointed: **the generation was claimed too late in the transaction.** A settings application propagates first and launches warm-ups second, so a stale acquire holding the pool lock in the gap between those two is still admitted — and installs a connection propagation has already walked past. When that application is the one that *removed* the server, nothing evicts it, and it survives until some later application happens to run.

Fixed by claiming the generation at the **start** of the application (`supersede_force_start()`, before `propagate_settings`), which makes the whole transaction a no-admittance window instead of just its tail. Claiming again in the warm-up pass is harmless, since only the newest value admits anything.

---

Status of the round: Greptile 5/5 clean, Qodo's one open item is its own earlier finding now marked resolved, and this was CodeRabbit's only finding.

@coderabbitai review
@greptileai

atusy:
Round 4 pushed as `cbe691e31`.

Addressed CodeRabbit's remaining test-gap finding with a deterministic settings-reload regression: the old detached warm-up is paused before admission, the reload is paused after propagation, and the stale acquire is released in that gap; the test verifies it cannot create a connection.

Validation: 3463 lib tests passed, clippy all targets/all features with `-D warnings` passed, and fmt/diff check are clean.

@coderabbitai review
@greptileai
/agentic_review
@codex review

atusy:
Round 5 pushed as `e5fed2d89` and `e445a0aaa`.

Fixed both actionable bot findings:
- oversized header-line errors now retain a bounded prefix of the rejected input, with a one-read regression test;
- separated the warm-up and retirement Rustdoc blocks so each documents the correct method.

Validation: 3463 lib tests passed, clippy all targets/all features with `-D warnings` passed, fmt and diff checks clean.

@coderabbitai review
@greptileai
/agentic_review
@codex review

atusy:
Round 6 pushed as `cb393ae5c`.

Addressed the remaining framing finding: a per-line ceiling error now prioritizes the rejected line prefix over an earlier stray header, with a regression covering a prior complete header followed by one oversized newline-free line.

Validation: 3464 lib tests passed, clippy all targets/all features with `-D warnings` passed, fmt and diff checks clean.

@coderabbitai review
@greptileai
/agentic_review
@codex review

atusy:
Round 7 pushed as `45886abf3`.

Addressed the split-read framing finding: oversized-line diagnostics now combine a bounded prefix of the accumulated line with the final rejected fragment, and a regression covers crossing the ceiling in a later absorb call after an earlier stray header.

Validation: 3465 lib tests passed, clippy all targets/all features with `-D warnings` passed, fmt and diff checks clean.

@coderabbitai review
@greptileai
/agentic_review
@codex review

PR #1048, opened 2026-09-02
fix(lsp): judge codeLens/resolve freshness on the region's effective offset

### LINKED ISSUE #1044: codeLens/resolve: reuse resolve_region_offset so lenses resolve inside #offset!/#trim! regions
## Summary

`codeLens/resolve` (#1023) and `documentLink/resolve` (#1025) both verify that the enveloped region is still fresh before dispatching to the producing server, but they compute freshness differently, and only the document-link version is right inside `#offset!` / `#trim!` regions.

- `Kakehashi::code_lens_region_is_fresh` (`src/lsp/lsp_impl/text_document/code_lens.rs`) takes the raw start byte from `node_tracker().lookup_position` and compares it against `envelope.offset`. In a region whose effective start was moved by `#offset!` or trimmed by `#trim!` (frontmatter, blockquote prose, …), the raw node start never equals the effective offset the lens was enveloped with, so every lens in such a region is reported stale and returned unresolved. The doc comment on that function records this as a known gap ("lenses stay unresolved … revisit if one appears").
- `Kakehashi::document_link_region_is_fresh` (`src/lsp/lsp_impl/text_document/document_link.rs`) re-runs injection resolution through `resolve_region_offset` and compares the full effective `RegionOffset` (per-line offsets included). Document links therefore resolve correctly inside those regions.

## Proposal

Port the document-link approach to code lenses: replace the raw-byte comparison in `code_lens_region_is_fresh` with `resolve_region_offset(...)` + `RegionOffset` equality, exactly as `document_link_region_is_fresh` does. The envelope already carries the effective offset, so no wire change is needed.

## Acceptance

- A lens produced inside a `#offset!` region (e.g. a markdown frontmatter injection) resolves instead of coming back unresolved.
- Existing `e2e_code_lens_resolve` cases (stale region, replaced producer, cancel) keep passing.
- Both freshness checks share one implementation, so the two resolve paths cannot drift again.

Context: found while reviewing #1025, where the document-link check was fixed in `83ae802c1` in response to a review finding; the same fix was deliberately not back-ported to code lenses in that PR to keep its scope.

https://claude.ai/code/session_01NhEfVnacTxTiURX5ftDwmn

atusy:
## Summary

`codeLens/resolve` reported every lens inside an `#offset!` / `#trim!` region (markdown YAML/TOML frontmatter, blockquote prose) as stale and returned it unresolved: its freshness gate compared the envelope's **effective** offset snapshot against the **raw** start byte of the tracked content node, and the two never agree once a directive shifts the region. `documentLink/resolve` already compared effective geometry (re-resolving the region through `resolve_region_offset`) and was correct.

Closes #1044

## What changed

- **One freshness gate.** `Kakehashi::region_offset_is_fresh(host_uri, region_id, offset)` lives next to `resolve_region_offset` and is the only resolve-side "is this region still where the envelope says" check; `codeLens/resolve` and `documentLink/resolve` both call it. It compares the whole effective `RegionOffset` (start line plus every per-line column offset), so an interior blockquote-prefix edit is also caught for lenses now, not just a moved start.
- **codeAction/resolve rebuilds its region the same way.** Its private copy of the snapshot → language → query → resolver → offset/end derivation is replaced by `resolve_region_offset`; the contiguity and whole-offset checks are unchanged.
- **Docs.** The "known limitation: frontmatter lenses/actions always fail soft" paragraphs on the code-lens and code-action gates are gone. The code-action one was already false (its live resolution was `#offset!`-adjusted all along); the region_offset module doc now lists the resolve gates among its callers.
- Unused `CodeLensEnvelope` / `DocumentLinkEnvelope` / `EnvelopeOffset` re-exports adjusted accordingly.

## Tests

- `e2e_code_lens_resolve_accepts_offset_frontmatter_region` (RED on main: lens at host line 1 comes back command-less; GREEN here: `mock resolved:lens-1`).
- `lazy_action_resolves_inside_offset_frontmatter_region` pins the code-action gate's frontmatter behaviour so it cannot regress to a raw-node comparison.
- Existing `e2e_code_lens_resolve` cases (shifted region fails soft, replaced producer, closed incarnation, cancel) and the document-link/code-action suites unchanged and passing; full e2e suite green.

https://claude.ai/code/session_01NhEfVnacTxTiURX5ftDwmn

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resolution of code actions and code lenses in offset-adjusted regions, including Markdown frontmatter.
  * Added consistent freshness checks for code lenses and document links to prevent stale results from being applied.
  * Fixed cases where valid actions in offset regions were incorrectly rejected during resolution.

* **Tests**
  * Added end-to-end coverage for resolving code actions and code lenses in offset-adjusted regions.

### DIFF
--- a/src/lsp/bridge.rs
+++ b/src/lsp/bridge.rs
@@ -83,12 +83,13 @@ pub(crate) use protocol::workspace_edit_preserves_line_prefixes;
 pub(crate) use protocol::workspace_edit_within_region;
 pub(crate) use text_document::host::{HostDocument, HostTextReader, normalize_host_goto_result};
 pub(crate) use text_document::{
-    CodeActionEnvelope, CodeLensEnvelope, DocumentLinkEnvelope, UpstreamCodeActionCaps,
-    bridge_code_actions, envelope_host_code_lenses, envelope_host_document_links,
-    extract_code_action_envelope, extract_code_lens_envelope, extract_document_link_envelope,
-    parse_code_actions_leniently,
+    CodeActionEnvelope, CodeLensEnvelope, UpstreamCodeActionCaps, bridge_code_actions,
+    envelope_host_code_lenses, envelope_host_document_links, extract_code_action_envelope,
+    extract_code_lens_envelope, extract_document_link_envelope, parse_code_actions_leniently,
 };
-pub(crate) use text_document::{KakehashiEnvelope, bridge_host_completion_items, extract_envelope};
+pub(crate) use text_document::{
+    EnvelopeOffset, KakehashiEnvelope, bridge_host_completion_items, extract_envelope,
+};
 pub(crate) use text_document::{OpenExpectation, OpenOutcome};
 pub(crate) use workspace::WorkspaceFolderSet;
 
--- a/src/lsp/bridge.rs
+++ b/src/lsp/bridge.rs
@@ -83,9 +83,9 @@ pub(crate) use protocol::workspace_edit_preserves_line_prefixes;
 pub(crate) use protocol::workspace_edit_within_region;
 pub(crate) use text_document::host::{HostDocument, HostTextReader, normalize_host_goto_result};
 pub(crate) use text_document::{
-    CodeActionEnvelope, CodeLensEnvelope, UpstreamCodeActionCaps, bridge_code_actions,
-    envelope_host_code_lenses, envelope_host_document_links, extract_code_action_envelope,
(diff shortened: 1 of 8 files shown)


atusy:
@coderabbitai review

atusy:
@greptileai

atusy:
@codex review

PR #1082, opened 2026-09-13
feat(language): support dedented injection documents

joshan-kana:
## Summary

Add `#set! injection.dedent` for bridge virtual documents. When enabled, Kakehashi removes the common leading spaces/tabs from non-blank lines after injection transforms and extends the per-line source offsets so downstream locations still map back to the original host document.

Blank lines do not constrain the common indent. Existing injections are unchanged unless the query opts in.

The virtual-document ADR documents the directive, and tests cover spaces, tabs, CRLF, blank lines, no-op input, and composition with runtime `#offset!` transforms.

## Validation

- `make check`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for the query-authored `#set! injection.dedent` directive in injected content.
  * Common leading spaces and tabs are removed while preserving line endings and accurate source-position mapping.
  * Blank lines do not affect indentation detection, and dedenting works for both individual and combined injected content.

* **Documentation**
  * Updated architecture documentation to describe dedent support and position-mapping behavior.


atusy:
Can you provide a concrete example that you aim to solve?

joshan-kana:
Thanks for having a look mate!

A concrete use case is embedded shell code in an indented Nix string, for example:

```nix
# bash
''
  if test -n "$HOME"; then
    echo "$HOME"
  fi
''
```

The Bash language server should ideally see the content without the Nix indentation:

```bash
if test -n "$HOME"; then
  echo "$HOME"
fi
```

That’s the main case I’m trying to support with `injection.dedent`.

My overarching goal is to have excellent LSP support for embedded languages within Nix files.

atusy:
@joshan-kana 
Understood.
On the other hand, I think this should rather be solved by parser and queries in https://github.com/nix-community/tree-sitter-nix .

Markdown parser/queries are solving the issue on their side.

`````markdown
- codeblock within bullet list

  ```python
  print(1)
  ```
`````

atusy:
Sorry, i noticed that I force pushed to your PRs by accident...

joshan-kana:
I looked into the tree-sitter-nix route a bit more.

Its injection query captures `string_fragment`, and the external scanner currently emits an indented string fragment as one contiguous token which can span multiple lines. Nix's indented-string behaviour isn't a fixed trim: it removes the minimum leading spaces across all non-blank lines in the whole string.

I don't see a query-only way to express that. A fixed range offset can trim an edge, but it can't calculate the common indentation and then remove that many columns from every physical line while preserving the host mapping. Kakehashi's transformed capture text also stays metadata-only and isn't substituted into bridge virtual documents.

Doing it entirely in tree-sitter-nix looks like it would require changing the grammar/scanner to expose indentation separately per line, while still somehow knowing the minimum indentation of the complete string before the earlier lines are represented.

If you had a particular parser/query structure in mind, I'm happy to try that approach. Otherwise this seems more naturally like a virtual-document transform, since the parser still needs source-accurate ranges while the downstream server needs the evaluated string layout.

atusy:
Thanks a lot for the explanation.
After some investigations on my side as well, I now agree with adding dedent feature.
Let me take some time "how" to support it.
It might be better to support by config (kakehashi.toml) rather than query.

atusy:
At least, `@injection.dedent` is not a good query name because it misleads dedent affects all injection-**related** features (e.g., tree-sitter highlighting)

PR #1083, opened 2026-09-13
feat(lsp): support protected gaps in combined injections

joshan-kana:
## Summary

Add explicit protected gaps for combined injection documents. Queries can capture host-owned syntax as `@injection.gap` and provide `injection.gap-placeholder`, with optional multiline prefix/suffix text, so downstream servers receive syntactically useful virtual content without gaining edit access to the host-owned span.

Protected host ranges are carried through completion and `completionItem/resolve`. Completion remains blocked for unsupported non-contiguous cases, and stale resolve geometry is rejected. Formatting projects downstream changes back onto editable spans, discards changes confined to synthetic gap text, and remains fail-closed for edits that cross protected syntax except whitespace-only formatter cleanup.

Ordinary `injection.combined` behavior is unchanged when no protected gap is configured.

## Validation

- `make check`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added support for explicit gaps when combining injected language regions, with configurable placeholders and surrounding text.
  - Completions now work across supported non-contiguous injected regions while preserving host-language syntax.
  - Formatting can process supported regions and safely project results back into the original document.

- **Bug Fixes**
  - Prevented completion and formatting edits from modifying protected host-language gaps.
  - Added safeguards against edits crossing protected syntax boundaries.
  - Prevented stale completion-resolution responses from retired language servers.


atusy:
@joshan-kana 

Thanks for the PR!
Could you share a concrete example of the syntax you want to support, along with the injection query and expected virtual document?

The appropriate placeholder seems to depend on both the language and the syntactic context. A real use case would help us understand the intended scope and assess whether the added complexity is justified.

joshan-kana:
The main use case is Nix interpolation inside an embedded language, for example:

```nix
# python
''
  result = ${someNixExpression}
  print(result)
''
```

If the interpolation is only replaced with whitespace, Python sees invalid syntax like `result = `. The idea is for the injection query to choose an appropriate placeholder, e.g. `None`, so the virtual document becomes:

```python
result = None
print(result)
```

while `${someNixExpression}` remains owned by Nix and cannot be edited by the Python language server.

I agree the placeholder depends on the language/context, which is why I made it query-defined rather than something Kakehashi chooses itself.

atusy:
Thanks for explaining the use case. For scripts complex enough to benefit from LSP support, I think passing values through environment variables, arguments, or configuration files would be a more intuitive approach. That would keep the embedded code syntactically complete and avoid maintaining injection queries with context-specific placeholders.

Given the implementation complexity and the potential edge cases around preserving host syntax during completion and formatting, I’m leaning toward not supporting protected gaps in Kakehashi.

Is there a concrete use case where those alternatives would be impractical, and interpolation within the embedded code is essential? I’d like to understand that before making a final decision.

joshan-kana:
Yep, there's a better example than the Python one I gave earlier: generated config.

```nix
# json
''
  {
    "port": ${toString cfg.port}
  }
''
```

This isn't a script with a runtime environment, so environment variables or arguments aren't an alternative. Nix is producing the final JSON text and the interpolation has to appear at that syntax position.

If the interpolation is only masked with whitespace, the JSON server effectively sees:

```json
{
  "port":
}
```

which is invalid. A query-defined placeholder such as `0` or `null` would instead keep the virtual document parseable while the original `${...}` range remains owned by Nix.

The same issue applies to generated JSON/YAML/TOML and other config/source snippets, not only shell scripts. Interpolation is part of how Nix constructs those files, so moving the value to a runtime channel isn't always possible.

I agree the current implementation is fairly large. If this use case is in scope, I'm happy to narrow the PR to the minimum protected-gap primitive and keep unsupported completion/formatting cases fail-closed rather than trying to make every non-contiguous edit shape work.

PR #1084, opened 2026-09-13
feat(cli): collect push-only diagnostics in diagnose

joshan-kana:
## Summary

Make `kakehashi diagnose` account for push-only diagnostic servers after `didOpen` instead of immediately racing into the pull-diagnostic pass.

For selected diagnostic fan-out entries that use `pushFallback` and do not advertise `textDocument/diagnostic`, the CLI now gives the existing diagnostic cache a bounded quiet period to settle. Cache activity restarts that quiet window; the total wait remains capped by the existing readiness timeout. A silent server is not treated as an operational failure because LSP exposes no push-completion capability.

The subsequent diagnostic request still uses the existing pull path and `pushFallback` merge. E2E coverage includes delayed pushes, explicit empty pushes, priority selection, and silent push-capable candidates.

## Validation

- `make check`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * `kakehashi diagnose` now includes diagnostics published asynchronously by downstream servers.
  * Push-only diagnostic results are collected after a brief settling period, including delayed results and explicit clean results.
  * Diagnostics from embedded documents are translated to host-document locations.

* **Bug Fixes**
  * Silent diagnostic servers no longer cause failures or generate false diagnostics.
  * Diagnostic filtering respects configured server priorities and participation rules.


atusy:
thanks, but i am not planning to support push-only diagnostics in diagnose for now.

Reasons:

* timing issue breaks reproducibility
* timing issue also makes it difficult to distinguish if the successful mean no-available diagnostics or no-arrived diagnostics.

atusy:
i would recommend open PRs to add textDocument/diagnostic to servers missing it

PR #1087, opened 2026-09-16
fix(nix): use stdenv.hostPlatform.isDarwin

kawarimidoll:
nixpkgs now wraps the `stdenv.is*` shorthands in `lib.warn` (NixOS/nixpkgs#518407), so evaluating the package or dev shell prints "stdenv.isDarwin is deprecated".

The warning also reaches consumers that install kakehashi through the flake, e.g. home-manager switch.

`stdenv.isDarwin` was only an alias for `hostPlatform.isDarwin`, so the selected inputs are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved platform detection for Darwin environments in development shells and package builds.


atusy:
thanks!

